PR #33, opened 2022-07-19
CU-2f7j85w Submodule jdbc

pablogthuillier:
This PR adds a new submodule for supporting jdbc using the capabilities from Scala-FX, like Control and Loom virtual threads fibers automatically yield on blocking IO calls.

Implements:
- ReadOnly operations
- LocalTransactions operations

This PR aims for the following properties:

Given I have a postgres database
And I write a program in scala-fx
When I run a query in DBReadOnly
Then my read only query runs in a db transaction and returns deserialized data correctly.

Given I have a postgres database
And I write a program in scala-fx
When I run a non-read-only query in ReadOnlyDBTransaction effect context
Then my query should fail and shift control to Errors[TransactionException].

Given I have a postgres database
And I write a program in scala-fx
When I run a set of operations including database queries in a LocalTransaction effect
And the database returns an error
Then the database state should roll back to before when the transaction occurred.

Given I have a postgres database
And I write a program in scala-fx
When I run a set of operations including database queries in a LocalTransaction effect
And the program shifts control to the Errors effect
Then the database state should roll back to before when the transaction occured.

Given I have a postgres database
And I write a program in scala-fx
When I run a set of operations including database queries in a LocalTransaction effect
And the database calls and program calls within the transaction succeeds
Then I should return the return type for the block and exit the transaction.

Given I have a postgres database
And I write a program in scala-fx
When I run a set of operations in any Transaction type
Then the rest of the independent scala FX code on structured threads should continue to run without being blocked by the library's blocking IO.

### DIFF
--- a/scalike-jdbc-scala-fx/src/main/scala/fx/DatabaseScala.scala
+++ b/scalike-jdbc-scala-fx/src/main/scala/fx/DatabaseScala.scala
@@ -6,17 +6,17 @@ import scalikejdbc.DB.{readOnly, CPContext, NoCPContext}
 import java.sql.SQLException
 import fx.ensure
 
-type Database[A] = (Structured, Control[SQLException]) ?=> Fiber[A]
-type Transaction[A] = (Structured, Control[Exception]) ?=> Fiber[A]
+type Database[A] = Control[SQLException] ?=> A
+type Transaction[A] = Control[Exception] ?=> A
 
 extension (db: DB.type)
   def readOnlyWithControl[A](execution: DBSession => A)(
       using context: CPContext = NoCPContext,
       settings: SettingsProvider = SettingsProvider.default): Database[A] =
-    fork { () => handle(DB.readOnly(execution))((e: SQLException) => e.shift) }
+    handle(DB.readOnly(execution))((e: SQLException) => e.shift)
 
   def localTransaction[A](execution: DBSession => A)(
       using context: CPContext = NoCPContext,
       boundary: TxBoundary[A] = TxBoundary.Exception.exceptionTxBoundary[A],
       settings: SettingsProvider = SettingsProvider.default): Transaction[A] =
-    fork { () => handle(DB.localTx(execution))((e: Exception) => e.shift) }
+    handle(DB.localTx(execution))((e: Exception) => e.shift)
--- a/scalike-jdbc-scala-fx/src/main/scala/fx/DatabaseScala.scala
+++ b/scalike-jdbc-scala-fx/src/main/scala/fx/DatabaseScala.scala
@@ -13,16 +13,10 @@ extension (db: DB.type)
   def readOnlyWithControl[A](execution: DBSession => A)(
       using context: CPContext = NoCPContext,
       settings: SettingsProvider = SettingsProvider.default): Database[A] =
-    fork { () =>
-      try DB.readOnly(execution)
-      catch case e: SQLException => e.shift
-    }
+    fork { () => handle(DB.readOnly(execution))((e: SQLException) => e.shift) }
 
   def localTransaction[A](execution: DBSession => A)(
-      implicit context: CPContext = NoCPContext,
+      using context: CPContext = NoCPContext,
       boundary: TxBoundary[A] = TxBoundary.Exception.exceptionTxBoundary[A],
       settings: SettingsProvider = SettingsProvider.default): Transaction[A] =
-    fork { () =>
-      try DB.localTx(execution)
-      catch case e: Exception => e.shift
-    }
+    fork { () => handle(DB.localTx(execution))((e: Exception) => e.shift) }
--- a/scalike-jdbc-scala-fx/src/main/scala/fx/DatabaseScala.scala
+++ b/scalike-jdbc-scala-fx/src/main/scala/fx/DatabaseScala.scala
@@ -1,0 +1,20 @@
+package fx
+
+import scalikejdbc.*
+import scalikejdbc.DB.{CPContext, NoCPContext, readOnly}
+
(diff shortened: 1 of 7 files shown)


scalike-jdbc-scala-fx/src/test/scala/fx/DatabaseSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+  testFX("Read only shift when tries to execute a non-read only operation") {
+    val names: Database[Int] = DB.readOnlyWithControl { implicit session =>
+      sql"update emp set name = 'Never changed in a readonly operation' where id = 1"
+        .update
+        .apply()
+    }
+    assertFX(run(structured(names.join.toEither.isLeft)) match {
+      case b: Boolean => false
+      case _ => true
+    })
```
jackcviers, line 28:
```suggestion
    assertFX(toEither(structured(DB.readOnlyWithControl { case given DBSession =>
      sql"update emp set name = 'Never changed in a readonly operation' where id = 1"
        .update
        .apply()
    }).join).isLeft)
```

If we remove the structured as suggested above, it can be:

```scala
assertFX(toEither(DB.readOnlyWithControl { case given DBSession =>
      sql"update emp set name = 'Never changed in a readonly operation' where id = 1"
        .update
        .apply()
    }).isLeft)
```

We can use `case given` instead of the deprecated `implicit`

jackcviers (changes requested):
It looks really nice @pablogthuillier!

I think you've pointed out that we need an `assertShiftsFX` in the munit lib or we can use that new munit PR to address this.

What do you think about removing the `Fiber` and allowing users to choose?

We shouldn't need to use `run` and pattern match inside of `testFX`, except in very rare circumstances (like that `shift` test case I mentioned).

Let me know what you think of the suggestions. As far as the api design goes, I think it's really solid and simple. Which is exactly what we're going for.

As always, suggestions are just that. I do think we need to limit the pattern matching and run calls in the tests at a minimum, but others are non-mandatory.

scalike-jdbc-scala-fx/src/main/scala/fx/DatabaseScala.scala
```diff
@@ -0,0 +1,28 @@
+package fx
+
+import scalikejdbc.*
+import scalikejdbc.DB.{readOnly, CPContext, NoCPContext}
+
+import java.sql.SQLException
+import fx.ensure
+
+type Database[A] = (Structured, Control[SQLException]) ?=> Fiber[A]
+type Transaction[A] = (Structured, Control[Exception]) ?=> Fiber[A]
```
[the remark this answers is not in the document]
pablogthuillier: It makes total sense. I have updated the PR removing Fiber layer.

scalike-jdbc-scala-fx/src/main/scala/fx/DatabaseScala.scala
```diff
@@ -0,0 +1,22 @@
...
+import scalikejdbc.DB.{readOnly, CPContext, NoCPContext}
+
+import java.sql.SQLException
+import fx.ensure
+
+type Database[A] = Control[SQLException] ?=> A
+type Transaction[A] = Control[Exception] ?=> A
+
+extension (db: DB.type)
+  def readOnlyWithControl[A](execution: DBSession => A)(
```
jackcviers, line 13: @pablogthuillier I wonder, could we get a referential transparency test for these methods? I don't think they are eager outside of a `run` context, but `testFX` eagerly applies the run so it is unclear. Anything that shows that subsequent calls to a `Database[A]` evaluate every time. What I'm getting at is that if it isn't RT, the `A` returns may have to be `() => A` or `Function0[A]` instead of just `A`.

Basically in:

```scala
val names1 = names // from the test
val names2 = names
```

names should evaluate twice. You could do that by updating a `var` counter or whatever. `Transaction` is easier to test because you can read afterwards to see that two items are inserted.

jackcviers (changes requested):
Rest of the changes are good to go. Just have the RT question. Thanks @pablogthuillier!

scalike-jdbc-scala-fx/src/test/scala/fx/DatabaseSpec.scala
```diff
@@ -0,0 +1,54 @@
+package fx
+
+import java.sql.SQLException
+import scalikejdbc.*
+import scalikejdbc.DB.*
+import java.util.concurrent.atomic.AtomicInteger
+
+class DatabaseSpec extends DatabaseSuite {
+
+  var counter: AtomicInteger = new AtomicInteger(0)
```
jackcviers, line 10:
```suggestion
  val counter: AtomicInteger = new AtomicInteger(0)
```
The AtomicInteger is mutable, so the reference doesn't have to be mutable.

scalike-jdbc-scala-fx/src/test/scala/fx/DatabaseSpec.scala
```diff
@@ -0,0 +1,54 @@
...
+  testFX("Read only shift when tries to execute a non-read only operation") {
+    assertFX(toEither(DB.readOnlyWithControl {
+      case given DBSession =>
+        sql"update emp set name = 'Never changed in a readonly operation' where id = 1"
+          .update
+          .apply()
+    }).isLeft)
+  }
+
+  var txCounter: AtomicInteger = new AtomicInteger(0)
```
jackcviers, line 31:
```suggestion
  val txCounter: AtomicInteger = new AtomicInteger(0)
```

jackcviers (approved):
Approve after those vars are vals

jackcviers (approved):

PR #34, opened 2022-07-19
Iteration over the MUnit integration

franciscodr:
So far,  as part of the MUnit integration into the encoding proposed in this project, @jackcviers presented a design that caught the exception thrown by the assertion methods when the condition wasn't satisfied. The exception was then safely handled using the Control capability to short-circuit the computation.

However, that design didn't cover the case where we want to test a computation that might fail with an error that doesn't extend Throwable. For example, this test would compile:
```
def myFunction(value1: Either[Throwable, Int], value2: Int): Errors[Throwable] ?=> Int =
  value1.bind + value2

testFX("The computation's result satisfies the assertion") {
  assertFX(myFunction(Left(Throwable("BOOM!")), 2) == 3)
}
```
but not this one:
```
def myFunction(value1: Either[String, Int], value2: Int): Errors[String] ?=> Int =
  value1.bind + value2

testFX("The computation's result satisfies the assertion") {
  assertFX(myFunction(Left("BOOM!"), 2) == 3)
}
```

This pull request proposes changes that will allow us to test computations that might short-circuit with a non-Throwable error.

### DIFF
--- a/munit-scalafx/src/main/scala/munit/fx/ScalaFXSuite.scala
+++ b/munit-scalafx/src/main/scala/munit/fx/ScalaFXSuite.scala
@@ -5,6 +5,7 @@ import _root_.fx._
 import org.junit.AssumptionViolatedException
 
 import scala.annotation.targetName
+import scala.reflect.Typeable
 
 /**
  * Provides functionality for testing within a scala-fx context.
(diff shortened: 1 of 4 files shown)


munit-scalafx/src/main/scala/munit/fx/ScalaFXSuite.scala
```diff
@@ -32,12 +33,14 @@ abstract class ScalaFXSuite extends FunSuite, ScalaFxAssertions:
      * @return
      *   The result of the test when a location can be pulled from given scope.
      */
-    def testFX[R <: Throwable, B](name: String)(body: Errors[R] ?=> A => B): Location ?=> Unit =
+    def testFX[R, F <: AssertionError: Typeable](name: String)(
```
jackcviers, line 36: `TypeTest` (the thing aliased by this) is something I was unaware of. I like that it solves the generic pattern match issue without using `Manifest`, which is more complicated and has a wider interface. That's pretty cool.

jackcviers (approved):
That's pretty nice! And thanks for fixing my string typos 😝!

munit-scalafx/src/main/scala/munit/fx/ScalaFXSuite.scala
franciscodr: Thanks, Jack. Yeah, I didn't know about the existence of `TypeTest` and `Typeable` either. I ran into them while reading the Scala 3 documentation looking for alternatives to deal with generic types and the type erasure in the JVM 🙂

jackcviers (approved):
Looks like it infers pretty well as these additional type parameters don't require any additional test changes!

PR #36, opened 2022-08-09
Design Document Scala Suspend

nomisRev:
Some thoughts and comments based on discussion with Raul, and compiler-plugin branch

### DIFF
--- a/Design.MD
+++ b/Design.MD
@@ -1,0 +1,85 @@
+## Use cases
+### Async programming
+### Generators
+## Implementation
+### Api
+#### Public
+
+```scala
+trait Continuation[A]
+  val context: ExecutionContext
+  def resume(value: A): Unit
+  def resumeWithException(error: Throwable): Unit
+```
+
+#### Internal
+
+Some notes studying the Kotlin implementation below.
+
+##### Continuation API
+
+Continuation has two methods, `resume` and `resumeWithException`.
+The first one is to return a _suspended value_,
+and the latter is to _bridge_ an async exception to caller thread.
+Kotlin uses `Result<A>` to merge both methods into 1, but it causes additional overhead and complexity.
+
+```scala
+trait Continuation[A]
+  def resume(value: A): Unit
+  def resumeWithException(error: Throwable): Unit
+```
+
+##### Interception
+
+Kotlin uses `ContinuationIntercepted`, and `Continuation#intercepted` for dispatching.
+The reason for this is that `Continuation` exists in Kotlin Std, but the concurrency implementation lives inside `KotlinX Coroutines`.
+The Kotlin Standard Library has _no concept_ of concurrency **at all**, so no abstraction such as `ExecutionContext` is available.
+
+Using `ExecutionContext` for the Scala implementation makes more sense, so we can replace `CoroutineContext` with `ExecutionContext`.
+`CoroutineContext` is further utilised to implement things such as `CoroutineContextLocal` alla `ThreadContextLocal`,
+and storing other information inside such as `CoroutineName` for debugging purposes so we can still maintain some `State` in a `CoroutineContext` like type.
+We will however not use it for _dispatching_ or scheduling for that we prefer to use `ExecutionContext` in Scala.
+
+```scala
+trait Continuation[A]
+  val context: ExecutionContext
+```
+
+##### Coroutine Builders
+
+Kotlin splits up the builders in `suspendCoroutine`, and `suspendCoroutineCancellable`.
+**Again** because Kotlin Std has no notion of concurrency, and thus no-notion of cancellation.
+We should probably not make this distinction/split in Scala, and prefer to have `suspendCoroutineCancellable` only.
+
+The coroutine builders should still rely on `SafeContinuation` and `intercepted`:
+1. `SafeContinuation` is a way of racing "foreign wrapping code" with the coroutine state machine.
+   When wrapping foreign code, SDKs often implicitly switch Thread, and thus we need to race to figure out if we had an _immediate_ or `COROUTINE_SUSPENDED` return.
+2. For _safety,_ we should _preserve the caller context_, and thus jump back to it. Otherwise, we also risk occupying SDK threads.
+
+=> SafeContinuation uses a `Undecided, Suspended & Resumed` enum to track state in Kotlin,
+but in Scala we can just use `val result: T | Undecided | Continuation.Suspended = Undediced`.
+The `Resumed` one is not needed, and the `Continuation.Suspended` already exists.
+
+Low-level builders (TBD).
+
+- suspendCoroutineUninterceptedOrReturn `val p1: completion`
+- startCoroutineUninterceptedOrReturn `(this as Function<Continuation<Any?>, Any?>).invoke(completion)`
+- createCoroutineUnintercepted (do we really need this?)
+
+For efficiency, we want to have `suspendCoroutineUninterceptedOrReturn` so you can return immediate values from `val x: Suspend %=> Int = 1`.
+This way the CPS compiler can also optimise these functions, and make them perform as immediate/regular functions inside the `Suspend` machinery.
+
+### Context functions
+### State machines
+### Resumption
+### CPS gen lambdas
+#### Tail call suspension / optimization
+#### Inlining and boxing
+### Spilling Variables
+#### Naming
+#### Clean up
+### Multi platform support
+#### Scala JVM
+#### Scala JS
+#### Scala Native
+### Debugging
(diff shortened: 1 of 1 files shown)


Design.MD
```diff
@@ -0,0 +1,85 @@
...
+### Async programming
+### Generators
+## Implementation
+### Api
+#### Public
+
+```scala
+trait Continuation[A]
+  val context: ExecutionContext
+  def resume(value: A): Unit
```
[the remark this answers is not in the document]
jackcviers: The reason here is that, unlike Threads, continuations capture their entire states. This allows for easy resource cleanup upon cancelation. Or scheduled cleanup.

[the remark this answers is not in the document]
jackcviers: We may want a shutdownHook method to implement that cleanup operation.

Design.MD
```diff
@@ -0,0 +1,85 @@
...
+```scala
+trait Continuation[A]
+  val context: ExecutionContext
+```
+
+##### Coroutine Builders
+
+Kotlin splits up the builders in `suspendCoroutine`, and `suspendCoroutineCancellable`.
+**Again** because Kotlin Std has no notion of concurrency, and thus no-notion of cancellation.
+We should probably not make this distinction/split in Scala, and prefer to have `suspendCoroutineCancellable` only.
```
raulraja, line 52: A side note here is that we need to consider our impl of `Structured` and see if we will support other strategies besides shutdown on failure like [ShutdownOnSuccess](https://download.java.net/java/early_access/loom/docs/api/jdk.incubator.concurrent/jdk/incubator/concurrent/StructuredTaskScope.ShutdownOnSuccess.html).
Also, how the cancellation wiring in these builders interacts with structured concurrency. Since we no longer depend on LOOM in the new model, we have to reimplement `Structured` or the equivalent of `coroutineScope`.

jackcviers, line 52: This is the call/cc implementation from classic continuations, yes?

Design.MD
```diff
@@ -0,0 +1,85 @@
...
+but in Scala we can just use `val result: T | Undecided | Continuation.Suspended = Undediced`.
+The `Resumed` one is not needed, and the `Continuation.Suspended` already exists.
+
+Low-level builders (TBD).
+
+- suspendCoroutineUninterceptedOrReturn `val p1: completion`
+- startCoroutineUninterceptedOrReturn `(this as Function<Continuation<Any?>, Any?>).invoke(completion)`
+- createCoroutineUnintercepted (do we really need this?)
+
+For efficiency, we want to have `suspendCoroutineUninterceptedOrReturn` so you can return immediate values from `val x: Suspend %=> Int = 1`.
```
raulraja, line 69:
```suggestion
For efficiency, we want to have `suspendCoroutineUninterceptedOrReturn` so you can return immediate values from `val x: Suspend ?=> Int = 1`.
```

Design.MD
```diff
@@ -0,0 +1,85 @@
...
+
+##### Coroutine Builders
+
+Kotlin splits up the builders in `suspendCoroutine`, and `suspendCoroutineCancellable`.
+**Again** because Kotlin Std has no notion of concurrency, and thus no-notion of cancellation.
+We should probably not make this distinction/split in Scala, and prefer to have `suspendCoroutineCancellable` only.
+
+The coroutine builders should still rely on `SafeContinuation` and `intercepted`:
+1. `SafeContinuation` is a way of racing "foreign wrapping code" with the coroutine state machine.
+   When wrapping foreign code, SDKs often implicitly switch Thread, and thus we need to race to figure out if we had an _immediate_ or `COROUTINE_SUSPENDED` return.
```
jackcviers, line 56: For immediate, we should be able to use the parasitic context.

raulraja (approved):
thanks @nomisRev !

jackcviers (commented):
Just some comments and questions @nomisRev

Design.MD
nomisRev: This function is like [IO.cancellable](https://typelevel.org/cats-effect/docs/2.x/datatypes/io#cancelable-processes) in the sense that it wraps a single function with cancellation mechanics.

I.e.

```kotlin
suspend fun sdkWrapper(): Int =
  suspendCoroutineCancellable { cont ->
     val callback = Callback { value, exception ->
       if(exception != null) cont.resumeWithException(exception)
       else cont.resume(value)
     }
     sdk.register(callback)
     CancelToken { sdk.unregister(callback) }
  }
```

So this is not related to structured concurrency perse, but we probably do want to have some `structured`/`coroutineScope` alternative with more strategies than KotlinX supports, and probably a saner default / syntax than Loom 😁

Design.MD
```diff
@@ -0,0 +1,85 @@
+## Use cases
+### Async programming
+### Generators
+## Implementation
+### Api
+#### Public
+
+```scala
+trait Continuation[A]
+  val context: ExecutionContext
```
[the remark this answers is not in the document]
nomisRev: A `Continuation` runs on a single "`Thread`" (`ExecutorService`/Context/pool), but it can interleave many other `Continuation`s running on other threads/pools.

So this `val context` represents the context this `Continuation` is running on,
if you run a `suspend () -> A` on a different context this requires you to instantiate a new `Continuation` with a new `val context`. This is useful so you can rely on `Continuation` to _preserve the context_.

Context preservation is something that we're discussing in our other comments, https://github.com/47deg/TBD/pull/36#discussion_r941384915. It's basically to prevent following scenarios.

```kotlin
val x: Suspend ?=> Int = suspendCoroutine { cont ->
  Thread({
    cont.resume(1) // resumes callback from `thread-1`
  }, "thread-1").start()
}

def main: Unit =
  println(Thread.currentThread().name) // global-context
  x()
  println(Thread.currentThread().name) // thread-1
```

Due to not preserving the context correctly within the `suspendCoroutine`, we are returned not on the caller thread but on `thread-1`.

This is unrelated to offering a optimised set of dispatchers like you say, which would be a great thing to have in the library. Wondering what your thoughts are after clarifying the reason.

Additional questions from my side:

 - Can we use `ExecutorService` in Scala JS?
 - Why does Scala have `ExecutionContext` ? (Especially if previous question was answered "yes").
 - Are there effect systems using `ExecutorService` rather than `ExecutionContext`?
 
 
IIRC CE2 relied on `ExecutionContext`, even `global` but also implemented a couple of their own. I.e. for trampolining / stack-safety.

Design.MD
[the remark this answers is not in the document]
nomisRev: Cancellation is indeed missing from this design document. We didn't get that far this morning. We can probably implement a similar system as CE3 in here.

I know the CE2 system can be build inside Kotlin Coroutines, so it should work here too.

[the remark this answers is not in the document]
nomisRev: ShutdownHook is IMO not something that we need to build into this, we can build it "around" it. See the example in Kotlin implemented for Node, Native and JS, https://github.com/arrow-kt/suspendapp.

https://github.com/arrow-kt/suspendapp/blob/700d8bc70e5f4a6a8e3906c7ece8557923fe2b36/src/jvmMain/kotlin/arrow/continuations/unsafe/unsafe.kt#L8

https://github.com/arrow-kt/suspendapp/blob/700d8bc70e5f4a6a8e3906c7ece8557923fe2b36/src/commonMain/kotlin/arrow/continuations/SuspendApp.kt#L36

Design.MD
nomisRev: Yes, this is the API to summon the `Continuation` and the ability to `resume`.

Design.MD
nomisRev: Discussed in the other comment. I am curious how this will play out in Scala. I think it can be done a lot simpler than the interception mechanism in Kotlin.

Design.MD
```diff
@@ -0,0 +1,85 @@
...
+We should probably not make this distinction/split in Scala, and prefer to have `suspendCoroutineCancellable` only.
+
+The coroutine builders should still rely on `SafeContinuation` and `intercepted`:
+1. `SafeContinuation` is a way of racing "foreign wrapping code" with the coroutine state machine.
+   When wrapping foreign code, SDKs often implicitly switch Thread, and thus we need to race to figure out if we had an _immediate_ or `COROUTINE_SUSPENDED` return.
+2. For _safety,_ we should _preserve the caller context_, and thus jump back to it. Otherwise, we also risk occupying SDK threads.
+
+=> SafeContinuation uses a `Undecided, Suspended & Resumed` enum to track state in Kotlin,
+but in Scala we can just use `val result: T | Undecided | Continuation.Suspended = Undediced`.
+The `Resumed` one is not needed, and the `Continuation.Suspended` already exists.
```
raulraja, line 61: `Resumed` is handled and set in the `Safe` continuation implementation which is part of the compiler plugin desugaring. For reference from the Kotlin impl port https://github.com/47deg/TBD/blob/43e5e220920930e6557159e0ed062912c759486a/continuationsPlugin/src/main/scala/continuations/SafeContinuation.scala

PR #39, opened 2022-10-05
Complete CPSExample

pabloconde:
- Completes CPSExample replicating an example which uses `await`.


.sdkmanrc
```diff
@@ -0,0 +1,4 @@
+# Enable auto-env through the sdkman_auto_env config
```
jackcviers, line 1: 👍🏻

jackcviers (approved):
Looks correct until we have a generater for the continuation. Until then it doesn't run, and I think that's what is expected here.

PR #40, opened 2022-10-07
Enable preview feature and update Java version

nychi713:
Updating Java version after Java19 was released past month.


.github/workflows/ci.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       - name: Setup Scala
         uses: olafurpg/setup-scala@v13
```
nomisRev, line 24: We need this Github Action, and use the more commonly used `actions/java-setup` See https://github.com/olafurpg/setup-scala/issues/49

Sbt 1.7.1 is pre-installed on [`ubuntu-latest`](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md), so we can simply use the `setup-java` action to install JDK 19.

```suggestion
        uses: actions/setup-java@v3.6.0
        with:
          distribution: 'temurin'
          java-version: 19
```

It's not documented well yet on the `actions/setup-java` but the `temurin` Java 19 version is available for all platforms except Windows according to this issue. https://github.com/adoptium/adoptium/issues/171

nomisRev: Same applies to all other workflows that rely on `setup-scala`

nomisRev:
Merged the commit with the flaky tests disabled, and green now 👍

nomisRev (approved):
LGTM 🙌

PR #49, opened 2022-10-22
circe fx

EstebanMarin:


jackcviers (commented):
Looks good.

circe-scalafx/src/main/scala/circefx/Circefx.scala
```diff
@@ -0,0 +1,26 @@
...
+
+def runProgram =
+  // TODO test with a Json
+  def decodeTest[A](c: io.circe.HCursor)(using d: Decoder[A], cx: Control[DecodingFailure]): A =
+    d(c).bind
+
+  def parserTest[A](s: String)(using p: Parser, cx: Control[Error], d: Decoder[A]) =
+    p.decode(s).bind
+
+  def transform[E, A](s: ValidatedNel[E, A]): Either[E, A] =
```
EstebanMarin, line 20: @jackcviers I am almost there as you see here. Now I am getting the following compilation erro
```
[error] -- [E007] Type Mismatch Error: /Users/estebanmarin/47deg/TBD/circe-scalafx/src/main/scala/circefx/Circefx.scala:21:47 
[error] 21 |    s.fold((e: NonEmptyList[E]) => Left(e.fold(identity)), (a: A) => Right(a))
[error]    |                                               ^^^^^^^^
[error]    |                                         Found:    Any => Any
[error]    |                                         Required: cats.kernel.Monoid[E]
```
which I can handle either by requesting evidence in the for of `def transform[E: A: Monoid]` but I contaminate upstream. So I think I need to provide an instance of `Monoid[E]` but not sure how

jackcviers: ValidatedNel already has a `.toEither`([toEither](https://www.javadoc.io/static/org.typelevel/cats-docs_2.13/2.8.0/cats/data/Validated.html#toEither:Either[E,A])) method, so we don't need to `fold` here to convert it to an `Either[NonEmptyList[E], A]`, which should be the return of this function (Not `Either[E, A]`):

```scala
import cats._
import cats.data._
import cats.implicits._

case class Foo(private val n: String)
val x: ValidatedNel[Foo, Int] = Foo("test").invalidNel[Int]
val y: Either[NonEmptyList[Foo], Int] = x.toEither
```

Where we might be getting confused is that we have an `Either[E, A]#bind` method that exists already that requires a `Control[E]` `using` parameter.

That `E` parameter can be anything. So to provide a `ValidatedNel[E, A]#bind` method, we need to remember that `ValidatedNel[E, A]` is a type alias for `Validated[NonEmptyList[E], A]` ([ValidatedNel](https://www.javadoc.io/doc/org.typelevel/cats-docs_2.13/latest/cats/data/index.html#ValidatedNel[+E,+A]=cats.data.Validated[cats.data.NonEmptyList[E],A]). So when we make our `bind` method for `ValidatedNel[E, A]` the `using` `Control`'s parameter is `NonEmptyList[E]` instead of just `E`.  Therefore, the signature for `ValidatedNel[E, A]#bind` will look like this:

```scala
//signature
extension[E, A](underlying: ValidatedNel[E, A]) def bind(using Control[NonEmptyList[E]]): A = ???
```
To implement the above, we need a method from `Validate[E, A] => A`. That method is [fold](https://www.javadoc.io/static/org.typelevel/cats-docs_2.13/2.8.0/cats/data/Validated.html#fold[B](fe:E=%3EB,fa:A=%3EB):B).

And a way to raise the `NonEmptyList[E]` in the error case in `fold`. We have that in our scala-fx library with [shift](https://github.com/47deg/TBD/blob/15d80a47ab70dd7221ee4b1d8b5316e07e2b3905/scala-fx/src/main/scala/fx/Continuation.scala#L18-L20).

Now we have everything to define the RHS of our `ValidatedNel[E, A]#bind` method signature from above.

Does that make sense @EstebanMarin?

EstebanMarin: @jackcviers it sure did, thanks a lot!

circe-scalafx/src/main/scala/circefx/Circefx.scala
```diff
@@ -0,0 +1,27 @@
...
+    d(c).bind
+
+  def parsing[A](s: String)(using p: Parser, cx: Control[Error], d: Decoder[A]): A =
+    p.decode(s).bind
+
+  object extensionTest:
+    def foldNonEmpty[E, A](x: NonEmptyList[E])(using control: Control[NonEmptyList[E]]): A =
+      // summon[Control[NonEmptyList[E]]]
+      ???
+
```
EstebanMarin, line 24: @jackcviers I noted on using [fold](https://www.javadoc.io/static/org.typelevel/cats-docs_2.13/2.8.0/cats/data/Validated.html#fold%5BB%5D(fe:E=%3EB,fa:A=%3EB):B)
now that leaves me with a sort of cast I need to do from `NonEmptyList[E]` to `A` as expressed in the `foldNonEmpty`. I am sure here I need to use the `shift`.

circe-scalafx/src/test/scala/fx/CirceSuite.scala
```diff
@@ -0,0 +1,30 @@
...
+
+class CirceSuite extends ScalaFXSuite, CirceFixtures:
+  combined.testFX("Circe FX test") {
+    case (x: Int, y: String) =>
+      assertEqualsFX(x, 1)
+  }
+
+  jsonError.testFX("JSON error") {
+    case (errorJson: String) =>
+      val test = parsing(errorJson)
```
EstebanMarin, line 28: @jackcviers I am struggling to provide evidence here and get the following
```
[error] -- Error: /Users/estebanmarin/47deg/TBD/circe-scalafx/src/test/scala/fx/CirceSuite.scala:28:35 
[error] 28 |      val test = parsing(errorJson)
[error]    |                                   ^
[error]    |no implicit argument of type io.circe.Parser was found for parameter p of method parsing in object Circefx
```

EstebanMarin:
@jackcviers I made some progress in the bind correctly (thanks for that), in preparation for our call. The function now parses a String to Json as expected. Moving forward I feel I need to:
- Make sure that I am formatting the file correctly. 
- Improve the assertion and make sure to test error handling
I am looking forward to our call.

PR #52, opened 2022-11-01
Plugin: Convert simple suspended def with no parameters and a suspendCoroutine

gregorath:
~it fails currently in runtime when creating `SafeContinuation` but opening the PR for visibility~

See commit [e6c3713](https://github.com/47deg/TBD/pull/52/commits/e6c37137a4242e0223d055db4bf919006b5ef45a) for comments on what each `val` step creates

### DIFF
--- a/continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
+++ b/continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
@@ -1,15 +1,15 @@
 package continuations
 
 import dotty.tools.dotc.ast.Trees.*
-import dotty.tools.dotc.ast.{tpd, Trees}
+import dotty.tools.dotc.ast.tpd
 import dotty.tools.dotc.core.Constants.Constant
 import dotty.tools.dotc.core.Contexts.{ctx, Context}
 import dotty.tools.dotc.core.Decorators.*
 import dotty.tools.dotc.core.Flags
 import dotty.tools.dotc.core.Names.termName
 import dotty.tools.dotc.core.StdNames.*
-import dotty.tools.dotc.core.Symbols.{Symbol, TermSymbol, *}
-import dotty.tools.dotc.core.Types.{AppliedType, OrType, Type}
+import dotty.tools.dotc.core.Symbols.*
+import dotty.tools.dotc.core.Types.{AppliedType, Type}
 import dotty.tools.dotc.plugins.{PluginPhase, StandardPlugin}
 import dotty.tools.dotc.report
 import dotty.tools.dotc.transform.{PickleQuotes, Staging}
@@ -50,24 +50,19 @@ class ContinuationsPhase extends PluginPhase:
     report.logWith("transformBlock")(tree.show)
     tree
 
-  var continuationTraitSym: ClassSymbol = _
-  var continuationObjectSym: Symbol = _
-  var safeContinuationClassSym: ClassSymbol = _
-  var interceptedMethodSym: TermSymbol = _
+  var defDefTransforms: DefDefTransforms = _
 
-  override def prepareForUnit(tree: Tree)(using Context): Context =
-    continuationTraitSym = requiredClass("continuations.Continuation")
-    continuationObjectSym = continuationTraitSym.companionModule
-
-    safeContinuationClassSym = requiredClass("continuations.SafeContinuation")
-
-    interceptedMethodSym =
+  override def prepareForDefDef(tree: DefDef)(using Context): Context =
+    defDefTransforms = new DefDefTransforms(
+      requiredClass("continuations.Continuation"),
+      requiredClass("continuations.SafeContinuation"),
       requiredPackage("continuations.intrinsics").requiredMethod("intercepted")
+    )
 
     ctx
 
   override def transformDefDef(tree: DefDef)(using Context): Tree =
-    transformSuspendNoParametersOneContinuationResume(tree)
+    defDefTransforms.transformSuspendContinuation(tree)
 
   @tailrec final def transformStatements(
       block: Block,
@@ -95,153 +90,3 @@ class ContinuationsPhase extends PluginPhase:
     tree match
       case Apply(_, _) => returnsContextFunctionWithSuspendType(tree)
       case _ => false
-
-  def hasOnlySuspendParam(tree: DefDef)(using Context): Boolean =
-    tree.termParamss match
-      case List(Nil, param :: Nil) => isSuspendType(param.tpe)
-      case _ => false
-
-  def getAllSubTrees(tree: Tree)(using Context): List[Tree] = {
-    val deepFolder: DeepFolder[ListBuffer[tpd.Tree]] =
-      new DeepFolder[ListBuffer[Tree]]((subTrees, tree) => subTrees += tree)
-
-    /*
-     * Needed for the `case Inlined(...)` because `DeepFolder.apply(...).foldOver` for the `Inlined` case  doesn't use
-     * the `call` tree and as a result the `Inlined` is not being unwrapped.
-     */
-    @tailrec
-    def recurse(trees: List[Tree], buf: ListBuffer[Tree]): List[Tree] =
-      trees match
-        case Nil =>
-          buf.toList
-        case Inlined(call, _, _) :: rest =>
-          recurse(call :: rest, buf)
-        case t :: rest =>
-          val (inlined, notInlined) =
-            deepFolder.apply(ListBuffer[Tree](), t).partition {
-              case _: Inlined => true
-              case _ => false
-            }
-          recurse(inlined.toList ++ rest, notInlined ++ buf)
-
-    recurse(tree.toList, ListBuffer[Tree]())
-  }
-
-  /*
-   * It works with only one top level `suspendContinuation` that just calls `resume`
-   * and it replaces the parent method body keeping any rows before the `suspendContinuation` (but not ones after).
-   */
-  def transformSuspendNoParametersOneContinuationResume(tree: DefDef)(using Context): Tree = {
-    if (hasOnlySuspendParam(tree)) {
-      val suspendContinuationName = "continuations.Continuation.suspendContinuation"
-      val resumeName = "continuations.Continuation.resume"
-
-      val suspendContinuationResumeCall: List[Tree] =
-        getAllSubTrees(tree.rhs)
-          .collect {
-            case Apply(TypeApply(fun, _), List(arg))
-                if fun.symbol.showFullName == suspendContinuationName =>
-              arg
-          }
-          .flatMap(getAllSubTrees)
-          .collect {
-            case Apply(fun, List(arg)) if fun.symbol.showFullName == resumeName =>
-              arg
-          }
-
-      if (suspendContinuationResumeCall.size == 1) {
-        val parent: Symbol = tree.symbol
-        val returnType: Type = tree.tpt.tpe
-        val resumeInput = suspendContinuationResumeCall.head
-
-        val rowsBeforeSuspend =
-          tree.rhs match
-            case Block(trees, _) =>
-              trees
-                .map {
-                  case Inlined(c, _, _) => c
-                  case t => t
-                }
-                .takeWhile(_.symbol.showFullName != suspendContinuationName)
-            case _ => List.empty[Tree]
-
-        val continuationTyped: Type =
-          continuationTraitSym.typeRef.appliedTo(returnType)
-
-        val completion =
-          newSymbol(parent, termName("completion"), Flags.LocalParam, continuationTyped)
-
-        val continuation1: ValDef =
-          ValDef(
-            newSymbol(parent, termName("continuation1"), Flags.Local, continuationTyped),
-            ref(completion))
-
-        val undecidedState =
-          ref(continuationObjectSym).select(termName("State")).select(termName("Undecided"))
-
-        val interceptedCall =
-          ref(interceptedMethodSym)
-            .appliedToType(returnType)
-            .appliedTo(ref(continuation1.symbol))
-            .appliedToNone
-
-        val safeContinuationConstructor =
-          New(ref(safeContinuationClassSym))
-            .select(nme.CONSTRUCTOR)
-            .appliedToType(returnType)
-            .appliedTo(interceptedCall, undecidedState)
-
-        val safeContinuation: ValDef =
-          ValDef(
-            newSymbol(
-              parent,
-              termName("safeContinuation"),
-              Flags.Local,
-              safeContinuationConstructor.tpe),
-            safeContinuationConstructor)
-
-        val safeContinuationRef =
-          ref(safeContinuation.symbol)
-
-        val suspendContinuation: ValDef =
-          ValDef(
-            newSymbol(
-              parent,
-              termName("suspendContinuation"),
-              Flags.Local,
-              ctx.definitions.IntType),
-            Literal(Constant(0))
-          )
-
-        val suspendContinuationResume =
-          safeContinuationRef.select(termName("resume")).appliedTo(resumeInput)
-
-        val suspendContinuationGetThrow =
-          safeContinuationRef.select(termName("getOrThrow")).appliedToNone
-
-        val body = Block(
-          rowsBeforeSuspend ++
-            (continuation1 :: safeContinuation :: suspendContinuation :: suspendContinuationResume :: Nil),
-          suspendContinuationGetThrow
-        )
-
-        val suspendedState =
-          ref(continuationObjectSym).select(termName("State")).select(termName("Suspended"))
-
-        /* TODO:
-         * Do we need to handle and remove `Continuation.State.Suspended.type` from this return type?
-         *
-         * If `Suspended` is actually being returned without being handled it will result to a `ClassCastException` as
-         * in runtime we have a method which expects a value that is of `returnType`.
-         */
-        // Any | Null | Continuation.State.Suspended.type
-        val finalMethodReturnType =
-          OrType(
-            OrType(ctx.definitions.AnyType, ctx.definitions.NullType, soft = false),
-            suspendedState.symbol.namedType,
-            soft = false)
-
-        DefDef(parent.asTerm, List(List(completion)), finalMethodReturnType, body)
-      } else tree
-    } else tree
-  }
--- a/continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
+++ b/continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
@@ -128,25 +128,24 @@ class ContinuationsPhase extends PluginPhase:
   }
 
   /*
-   * It works with only one `suspendContinuation` that just calls `resume`
-   * and it replaces the whole parent method body
+   * It works with only one top level `suspendContinuation` that just calls `resume`
+   * and it replaces the parent method body keeping any rows before the `suspendContinuation` (but not ones after).
    */
   def transformSuspendNoParametersOneContinuationResume(tree: DefDef)(using Context): Tree = {
     if (hasOnlySuspendParam(tree)) {
+      val suspendContinuationName = "continuations.Continuation.suspendContinuation"
+      val resumeName = "continuations.Continuation.resume"
 
       val suspendContinuationResumeCall: List[Tree] =
         getAllSubTrees(tree.rhs)
           .collect {
             case Apply(TypeApply(fun, _), List(arg))
-                if fun
-                  .symbol
-                  .showFullName == "continuations.Continuation.suspendContinuation" =>
+                if fun.symbol.showFullName == suspendContinuationName =>
               arg
           }
           .flatMap(getAllSubTrees)
           .collect {
-            case Apply(fun, List(arg))
-                if fun.symbol.showFullName == "continuations.Continuation.resume" =>
+            case Apply(fun, List(arg)) if fun.symbol.showFullName == resumeName =>
               arg
           }
 
@@ -154,6 +153,17 @@ class ContinuationsPhase extends PluginPhase:
         val parent: Symbol = tree.symbol
         val returnType: Type = tree.tpt.tpe
         val resumeInput = suspendContinuationResumeCall.head
+
+        val rowsBeforeSuspend =
+          tree.rhs match
+            case Block(trees, _) =>
+              trees
+                .map {
+                  case Inlined(c, _, _) => c
+                  case t => t
+                }
+                .takeWhile(_.symbol.showFullName != suspendContinuationName)
+            case _ => List.empty[Tree]
 
         val continuationTyped: Type =
           continuationTraitSym.typeRef.appliedTo(returnType)
@@ -210,7 +220,8 @@ class ContinuationsPhase extends PluginPhase:
           safeContinuationRef.select(termName("getOrThrow")).appliedToNone
 
         val body = Block(
-          continuation1 :: safeContinuation :: suspendContinuation :: suspendContinuationResume :: Nil,
+          rowsBeforeSuspend ++
+            (continuation1 :: safeContinuation :: suspendContinuation :: suspendContinuationResume :: Nil),
           suspendContinuationGetThrow
         )
 
--- a/continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
+++ b/continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
@@ -50,19 +50,8 @@ class ContinuationsPhase extends PluginPhase:
     report.logWith("transformBlock")(tree.show)
     tree
 
-  var defDefTransforms: DefDefTransforms = _
-
-  override def prepareForDefDef(tree: DefDef)(using Context): Context =
-    defDefTransforms = new DefDefTransforms(
-      requiredClass("continuations.Continuation"),
-      requiredClass("continuations.SafeContinuation"),
-      requiredPackage("continuations.intrinsics").requiredMethod("intercepted")
-    )
-
-    ctx
-
   override def transformDefDef(tree: DefDef)(using Context): Tree =
-    defDefTransforms.transformSuspendContinuation(tree)
+    new DefDefTransforms().transformSuspendContinuation(tree)
 
   @tailrec final def transformStatements(
       block: Block,
--- a/continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
+++ b/continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
@@ -52,14 +52,14 @@ class ContinuationsPhase extends PluginPhase:
 
   var continuationTraitSym: ClassSymbol = _
   var continuationObjectSym: Symbol = _
-  var safeContinuationClassApplySym: ClassSymbol = _
+  var safeContinuationClassSym: ClassSymbol = _
   var interceptedMethodSym: TermSymbol = _
 
   override def prepareForUnit(tree: Tree)(using Context): Context =
     continuationTraitSym = requiredClass("continuations.Continuation")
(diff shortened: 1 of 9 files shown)


gregorath:
so now this "works" but fails because the [SafeContintuation](https://github.com/47deg/TBD/blob/compiler-plugin/continuationsPlugin/src/main/scala/continuations/SafeContinuation.scala) does not work (error `Caused by: java.lang.IllegalAccessException: class continuations.SafeContinuation$ cannot access a member of class continuations.SafeContinuation with modifiers "private volatile"`)

But hardcoding a result value in `SafeContintuation` the transformation happens and the result is being returned in the [example](https://github.com/47deg/TBD/pull/52/files#diff-98d6a073ba084ca5d9135c3b60f115afa9a45962894ea59805459339100dac6cR61).

There is also [this bit](https://github.com/47deg/TBD/pull/52/files#diff-bfdffcfa12331080aaef50b40eae8ba34ac1bbfa58bb69c99d4ceef4b5607d7cR220) regarding the result from `SafeContintuation.getOrThrow`. In Kotlin it seems that the `Suspended` value goes away with [DebugProbes](https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/jvm/src/kotlin/coroutines/jvm/internal/DebugProbes.kt#L17) (?)

jackcviers:
> There is also [this bit](https://github.com/47deg/TBD/pull/52/files#diff-bfdffcfa12331080aaef50b40eae8ba34ac1bbfa58bb69c99d4ceef4b5607d7cR220) regarding the result from `SafeContintuation.getOrThrow`. In Kotlin it seems that the `Suspended` value goes away with [DebugProbes](https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/jvm/src/kotlin/coroutines/jvm/internal/DebugProbes.kt#L17) (?)

Maybe a question for @nomisRev?

nomisRev:
I'm not 100% sure I understand the problem statement here, but there seem to be two questions

1. >Caused by: java.lang.IllegalAccessException: class continuations.SafeContinuation$ cannot access a member of class continuations.SafeContinuation with modifiers "private volatile"

Where is this coming from exactly, can you share a stack-trace? If it's coming from [here](https://github.com/47deg/TBD/blob/d23cee318cd8e0639eec6b4ad849aeef6db80dff/continuationsPlugin/src/main/scala/continuations/SafeContinuation.scala#L36), that line seems redundant to me 🤔 Since no new _result_ is ever calculated.

2. > There is also [this bit](https://github.com/47deg/TBD/pull/52/files#diff-bfdffcfa12331080aaef50b40eae8ba34ac1bbfa58bb69c99d4ceef4b5607d7cR220) regarding the result from SafeContintuation.getOrThrow. In Kotlin it seems that the Suspended value goes away with [DebugProbes](https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/jvm/src/kotlin/coroutines/jvm/internal/DebugProbes.kt#L17) (?)


`DebugProbes` and `Continuation.State.Suspended` are unrelated. The `SafeContinuation` type that is needed to implement the "higher-level" `suspendCoroutine` function alla `IO.async`. It allows users to:
 -  avoid having to deal with `COROUTINE_SUSPENDED` or `Any | Null`
 - Re-schedule on the original dispatcher before returning, since this method is typically used form FFI like code and avoiding accidentally hogging a library dispatcher or pool.

`DebugProbes` interacts with 3 methods `probeCoroutineCreated`, `probeCoroutineSuspended`, and `probeCoroutineResumed`. This way `DebugProbes` can track all created `Coroutines` and track if they're in _running_ or _suspended_ state. Like a Fiber-dump would do.

continuationsPlugin/src/main/scala/continuations/SafeContinuation.scala
```diff
@@ -29,27 +28,20 @@ class SafeContinuation[-T](val delegate: Continuation[T], initialResult: Any | N
...
         return Continuation.State.Suspended
       result = this.result
     if (result == Continuation.State.Resumed) Continuation.State.Suspended
     else if (result.isInstanceOf[Left[_, _]])
       throw result.asInstanceOf[Left[Throwable, _]].value
-    else result
+    else if (result.isInstanceOf[Right[_, _]])
+      result.asInstanceOf[Right[_, _]].value
+    else
+      result // COROUTINE_SUSPENDED
```
gregorath, line 40: I believe this is right, as mentioned [here](https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/jvm/src/kotlin/coroutines/SafeContinuationJvm.kt#L59).

It is not quite clear to me whether this `COROUTINE_SUSPENDED` would be inside the `Right`, in which case this last `else` should throw something like `throw IllegalStateException("Unexpected result value" ++ result.toString)` or the `COROUTINE_SUSPENDED` would be the value of the result.

In either case I think it is fine to return the `result`, and re-evaluate when we have build all the functionality that handles the response from this `getOrThrow` method

nomisRev: > or the COROUTINE_SUSPENDED would be the actual value of the result.

It's the actual value of the result. In our case `Continuation.State.Suspended`. This can be coming from other surrounding suspending code, and needs to be passed through. It signals to the CPS machinery that at some later point `resumeWith` will be called, which will in turn call our `resumeWith` at a later point in time.

nomisRev: I found this non braced `if-else` code in Scala 3 very hard to read 😅 I guess I still need to get used to it. Perhaps using `if` statements it can be a bit better readable?

jackcviers:
Me too. Add it to the style guide card.

Returning suspend is intentional as Simon said. I haven't pulled the code
to run it yet. Will have some time this weekend.

Jack Viers - 515.313.7838 ***@***.***

continuationsPlugin/src/main/scala/continuations/SafeContinuation.scala
gregorath: > Perhaps using if statements it can be a bit better readable?

you mean like this 
```
    if (result == Continuation.State.Resumed) return Continuation.State.Suspended
    if (result.isInstanceOf[Left[_, _]]) throw result.asInstanceOf[Left[Throwable, _]].value
    if (result.isInstanceOf[Right[_, _]]) return result.asInstanceOf[Right[_, _]].value
```

to be honest in this case I find the use of `return` a bit weird in Scala (more than the `if-else`) 😂 
but happy to change it once confirmed this is what you meant

gregorath: >I also wonder if Either is the way to go here, or we should use union types here instead since we never rely on the bias. Not sure if union types are free of the allocation overhead. 

I think union types are unboxed but it looks like the choice of `Either` was done intentionally to distinguish the error from the values like [override def resume(result: Either[Throwable, Any | Null])](https://github.com/47deg/TBD/blob/23eaf30ce4a817ab457ff5b08888263d2174e520/continuationsPlugin/src/main/scala/continuations/jvm/internal/ContinuationImpl.scala#L11).

continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
```diff
@@ -74,3 +96,141 @@ class ContinuationsPhase extends PluginPhase:
     tree match
       case Apply(_, _) => returnsContextFunctionWithSuspendType(tree)
       case _ => false
+
+  def hasOnlySuspendParam(tree: DefDef)(using Context): Boolean =
+    tree.termParamss match
+      case List(Nil, param :: Nil) => isSuspendType(param.tpe)
```
jackcviers, line 101: This will work on *our* lone example. But it would fail if there are more using params, and it would accidentally trigger the rewrite in something like this:

```scala
def myNonSupsendFunction(s: Suspend): = ???
```

To check if tree is a `DefDef`, and if it has a `using Suspend` parameter, we need to check `tree.isDef` is `true` cast it as a `tpd.DefDef`, then find if a parameter in its `paramss` exists that is `Flags.Given` and that the parameter's type is `continuations.Suspend`.

Here's a working example from jackcviers/TBD:30m3gh3-non-suspending-0-arity local work progress with tests and results.

```scala
package continuations

import dotty.tools.dotc.ast.Trees
import dotty.tools.dotc.ast.tpd
import dotty.tools.dotc.core.Contexts.Context
import dotty.tools.dotc.core.Flags
import dotty.tools.dotc.report

object HasSuspendParameter {

  def unapply(tree: tpd.Tree)(using c: Context): Option[tpd.Tree] = Option(tree).filter{t =>
    t.isDef && t.asInstanceOf[tpd.DefDef].paramss.exists{ pc =>
      pc.exists{ v => v.denot.symbol.is(Flags.Given) && v.tpe.classSymbol.showFullName == "continuations.Suspend" }
    }
  }
}
```
Tests:
```scala
package continuations

import dotty.tools.dotc.core.Contexts.Context
import dotty.tools.dotc.core.Symbols
import dotty.tools.dotc.core.Types.Type
import dotty.tools.dotc.ast.tpd
import munit.FunSuite
import dotty.tools.dotc.core.Names
import dotty.tools.dotc.core.Flags
import dotty.tools.dotc.core.Constants.Constant.apply
import dotty.tools.dotc.core.Constants
import dotty.tools.dotc.ast.Trees

class HasSuspendParmeterSuite extends FunSuite, CompilerFixtures {

  compilerContextWithContinuationsPlugin.test(
    "#unapply(defDefTree): def mySuspend()(using Suspend): Int = 1 should be Some(tree)") {
    case given Context =>
      val c = summon[Context]
      val suspend = Symbols.requiredClass("continuations.Suspend")
      val d = tpd.DefDef(
        Symbols
          .newSymbol(
            c.owner,
            Names.termName("mySuspend"),
            Flags.EmptyFlags,
            c.definitions.IntType)
          .asTerm,
        List(
          List(),
          List(
            Symbols.newSymbol(
              c.owner,
              Names.termName("x$1"),
              Flags.union(Flags.GivenOrImplicit, Flags.Param),
              suspend.info))),
        c.definitions.IntType,
        tpd.Literal(Constants.Constant(1))
      )
      assertEquals(HasSuspendParameter.unapply(d), Some(d))
  }

  compilerContextWithContinuationsPlugin.test(
    "#unapply(defDefTree): def mySuspend(s: Suspend): Int = 1 should be None") {
    case given Context =>
      val c = summon[Context]
      val suspend = Symbols.requiredClass("continuations.Suspend")
      val d = tpd.DefDef(
        Symbols
          .newSymbol(
            c.owner,
            Names.termName("mySuspend"),
            Flags.EmptyFlags,
            c.definitions.IntType)
          .asTerm,
        List(
          List(
            Symbols.newSymbol(
              c.owner,
              Names.termName("s"),
              Flags.union(Flags.Param),
              suspend.info))),
        c.definitions.IntType,
        tpd.Literal(Constants.Constant(1))
      )
      assertEquals(HasSuspendParameter.unapply(d), None)
  }
}
```
Results:
```scala
continuations.HasSuspendParmeterSuite:
  + #unapply(defDefTree): def mySuspend()(using Suspend): Int = 1 should be Some(tree) 0.458s
  + #unapply(defDefTree): def mySuspend(s: Suspend): Int = 1 should be None 0.053s
[info] Passed: Total 2, Failed 0, Errors 0, Passed 2
[success] Total time: 2 s, completed Nov 4, 2022, 12:19:15 PM
```
.

continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
```diff
@@ -48,6 +50,26 @@ class ContinuationsPhase extends PluginPhase:
...
+    continuationTraitSym = requiredClass("continuations.Continuation")
+    continuationObjectSym = requiredModule("continuations.Continuation")
+
+    safeContinuationClassApplySym = requiredModule("continuations.SafeContinuation")
+      .requiredMethod("apply", List(defn.AnyType, defn.AnyType))
+
+    interceptedMethodSym =
+      requiredPackage("continuations.intrinsics").requiredMethod("intercepted")
+
+    ctx
```
jackcviers, line 67: ❓ I'm not an expert on the dotty compiler (yet), but is `prepareForUnit` where we want to do this work? Wouldn't we want to do this in `prepareForDefDef` since we're transforming a `DefDef` tree?

continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
```diff
@@ -48,6 +50,26 @@ class ContinuationsPhase extends PluginPhase:
     report.logWith("transformBlock")(tree.show)
     tree
 
+  var continuationTraitSym: ClassSymbol = _
+  var continuationObjectSym: TermSymbol = _
+  var safeContinuationClassApplySym: TermSymbol = _
+  var interceptedMethodSym: TermSymbol = _
```
jackcviers, line 56: 💡 I've been creating the transformations as a class `continuations.DefDefTransforms`. I think these could be class parameters and we could avoid the empty vars here if we did it that way.

continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
jackcviers: 💡 If we create a class like suggested above, we can move this method there (and avoid possible merge conflicts).

continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
```diff
@@ -74,3 +96,141 @@ class ContinuationsPhase extends PluginPhase:
...
+          recurse(call :: rest, buf)
+        case t :: rest =>
+          val (inlined, notInlined) =
+            deepFolder.apply(ListBuffer[Tree](), t).partition {
+              case _: Inlined => true
+              case _ => false
+            }
+          recurse(inlined.toList ++ rest, notInlined ++ buf)
+
+    recurse(tree.toList, ListBuffer[Tree]())
```
jackcviers, line 127: 💡 If we move the transformation to a class like suggested above, we can move this method there.

continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
```diff
@@ -74,3 +96,141 @@ class ContinuationsPhase extends PluginPhase:
...
+      case _ => false
+
+  def getAllSubTrees(tree: Tree)(using Context): List[Tree] = {
+    val deepFolder: DeepFolder[ListBuffer[tpd.Tree]] =
+      new DeepFolder[ListBuffer[Tree]]((subTrees, tree) => subTrees += tree)
+
+    /*
+     * Needed for `case Inlined` because DeepFolder.foldOver for Inlined doesn't use `call` tree so the Inlined
+     * is not being unwrapped (???)
+     */
```
jackcviers, line 111: Not a dotty expert (yet), but you may be able to avoid this by forcing the rhs tree of the DefDef, and get away with a `TreeCopier` [instead](https://github.com/lampepfl/dotty/blob/63344e7337727f886d3f7f1a0ad0234dd7670df8/compiler/src/dotty/tools/dotc/ast/Trees.scala#L1131-L1365) to selectively replace the statements in the tree via copying. I don't know if that's the standard way of doing it, but see the [LambdaLift](https://github.com/lampepfl/dotty/blob/a1157db1e44f23b1f1e605cf40ca2c0e88265c87/compiler/src/dotty/tools/dotc/transform/LambdaLift.scala#L192-L202) phase as an example.

continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
jackcviers: Additionally, since these are compiler-generated, would they be flagged `synthetics`?

jackcviers (changes requested):

continuationsPlugin/src/main/scala/continuations/SafeContinuation.scala
nomisRev: We can investigate it later with benchmarks, but this is being used in a hot-loop. Kotlin has a special type Result that avoids allocation in the non-Throwable case and pays a penalty for Throwable where it’s arguably neglectable.

But let’s post-pone that. It’s not important for functionality, and doing performance optimisations make more sense when backed by benchmarks.

jackcviers: > > Perhaps using if statements it can be a bit better readable?
> 
> you mean like this?
> ```
>     if (result == Continuation.State.Resumed) return Continuation.State.Suspended
>     if (result.isInstanceOf[Left[_, _]]) throw result.asInstanceOf[Left[Throwable, _]].value
>     if (result.isInstanceOf[Right[_, _]]) return result.asInstanceOf[Right[_, _]].value
>     result
> ```
> 
> to be honest in this case I find the use of `return` in Scala a bit weird  (more than the `if-else`) 😂 
> but happy to change it once confirmed this is what you meant

I think he just meant

```scala
If(){
  ...
} else {
  ...
} ...
```

nomisRev: Yes, that is exactly what I meant @jackcviers 👍

gregorath: added the braces in [1a5f9ef](https://github.com/47deg/TBD/pull/52/commits/1a5f9ef8011ed082a480c0da6afa4bae1cc71944)

continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
gregorath: added the more general case https://github.com/47deg/TBD/pull/52/commits/1a5f9ef8011ed082a480c0da6afa4bae1cc71944#diff-fade65ebf959155389570d62d274801c2f7607f965a273661e5f776636d6393fR131

continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
gregorath: yep you are right https://github.com/47deg/TBD/pull/52/commits/1a5f9ef8011ed082a480c0da6afa4bae1cc71944#diff-bfdffcfa12331080aaef50b40eae8ba34ac1bbfa58bb69c99d4ceef4b5607d7cR55

continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
gregorath: added the [class](https://github.com/47deg/TBD/pull/52/commits/1a5f9ef8011ed082a480c0da6afa4bae1cc71944#diff-fade65ebf959155389570d62d274801c2f7607f965a273661e5f776636d6393fR19) with parameters but we still need one [var](https://github.com/47deg/TBD/pull/52/commits/1a5f9ef8011ed082a480c0da6afa4bae1cc71944#diff-bfdffcfa12331080aaef50b40eae8ba34ac1bbfa58bb69c99d4ceef4b5607d7cR53). 
not sure if it is a problem but could create the class inside the `transformDefDef`

continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
gregorath: got rid of this method, it was helping us to deal with `Inlined` and also look for nested `suspendContinuation` calls. but I guess atm we don't need to worry about nested ones, just look for top level, and that simplified the code for now https://github.com/47deg/TBD/pull/52/commits/1a5f9ef8011ed082a480c0da6afa4bae1cc71944#diff-fade65ebf959155389570d62d274801c2f7607f965a273661e5f776636d6393fR142

continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
gregorath: as said above, we don't have this method anymore

gregorath:
@jackcviers all the comments have been addressed and did some more "improvements" (and felt the pain of manually writing the tree) in case you want to have a look. I guess the structure of the class and the grouping of methods will make more sense once we add more transformation cases

jackcviers:
I'll take a look tonight @gregorath

continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
jackcviers: Yes. Looks good.

continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
jackcviers: I'm not convinced we actually need those `requireX` lookups outside of the actual `DefDefTransforms` class for this use case. They live in the `Symbols` table of the compiler via loading from the classpath. So inside `DefDefTransforms.transformSuspendContinuation(tree)`, the first thing we could do is pull them in with:

```scala
val ContinuationClazz = requiredClass("continuations.Continuation")
val SafeContinuationsClazz = requiredClass("continuations.SafeContinuation")
val interceptedMethod = requiredPackage("continuations.intrinsics").requiredMethod("intercepted")
...rest of transformSuspendedContinuation(tree)
```
or something similar.

If other methods inside `DefDefTransforms` require the symbols, they can pull them from the `Symbols` object with `Symbols.requiredClass` and `Symbols.requiredPackage`, etc, in the same way. If that gets to be too repetitive, we can look them up in `ContinuationsPlugin.transformDefDef` and then pass them to `DefDefTransforms`'s constructor for reuse (assuming they don't get changed inside our transform).

#### Aside: Sharing mutable state within the phase

There is a `Store` ([Store](https://github.com/lampepfl/dotty/blob/1276ba54bd172c68e4ac733bcef31a4b141b8bc6/compiler/src/dotty/tools/dotc/util/Store.scala#L10)) for generic object storage on `Context`, so to share common values that may change between `transform<TreeType>` methods within the same phase we'd setup the initial state in the `prepare<TreeType>`, add them to the `Context#store` with `Context#addLocation[T](T):Location[T]`([addLocation](https://github.com/lampepfl/dotty/blob/b3f870fdf2f8bedd7d73af0cb80ece5142d2ccec/compiler/src/dotty/tools/dotc/core/Contexts.scala#L660)), and we'd `Promise#tryComplete(Location[T])` to store that in a `Promise[Location[T]]` `val` on the phase. `transform<TreeType>` methods and callees could then look them up with `Context#store#apply(Location[T])`([Store#apply](https://github.com/lampepfl/dotty/blob/1276ba54bd172c68e4ac733bcef31a4b141b8bc6/compiler/src/dotty/tools/dotc/util/Store.scala#L31)), where we get the values from the `Promise[Location[T]]` `val`s on the phase. If we need to mutate the value at `Location[T]`, we do so with `Context#updateStore[T](Location[T], T):Context`([Context#updateStore](https://github.com/lampepfl/dotty/blob/b3f870fdf2f8bedd7d73af0cb80ece5142d2ccec/compiler/src/dotty/tools/dotc/core/Contexts.scala#L672)), which sets the value at `Location[T]` on the context store and returns the context.  We're not doing more than one `transform<TreeType>` transform, yet, so we don't have this problem at this time, but we might need to store contextual values at some point, which is why I included this aside.

A sketch:

```scala
// in ContinuationsPluginPhase
private val myMutableInt = Promise[Int]

override def prepareForApply(tree: Apply)(using Context): Context = 
  val c = summon[Context]
  if(!myMutableInt.isCompleted){
    val _ = myMutableInt.tryComplete(c.addLocation(1))
  }
override def transformApply(tree: Apply)(using Context): Context = 
  new TransformApply(Await.result(myMutableInt.future, Duration.Inf))
  
// . . .
override def transformAssign(tree: Assign)(using Context): Tree =
  new TransformAssign(Await.result(myMutableInt.future, Duration.Inf))

// . . .

// in some method in TransformAssign, which stores the Location[Int] as loc via its constructor
val currentInt = context.store(loc)
val newInt = currentInt + 1
val _ = summon[Context].updateStore(loc, newInt)

// . . .

// in some method in  or TransformApply, which stores the Location[Int] as loc via its constructor
val currentInt = context.store(loc)
val newInt = currentInt + 1
val _ = summon[Context].updateStore(loc, newInt)

// . . .
```
We don't have to use promises and futures, but we need something that can only be completed once to do the setup (because we don't know the order `prepareFor<TypeTree>` will be called in), so it's `Promise` or `CompletableFuture`, I think, for that use case.

This lets us share the mutable `Int` between `transform<TypeTree>` calls and have a consistent state within a compilation unit, without using any mutable vars, and so that we're keeping contextual things in the `Context` where they belong, and allows us to operate on them without having to know the order of application of the transformations. There's also `Context#properties`, but apparently that is slower since `Context#store` is actually just an array, and a `Location[T]` is just a pointer to the index into that array.

jackcviers:
> @jackcviers all the comments have been addressed and did some more "improvements" (and felt the pain of manually writing the tree) in case you want to have a look. I guess the structure of the class and the grouping of methods will make more sense once we add more transformation cases

Only one comment that remains unaddressed. I don't think we need to store the transform class instance to avoid looking up the symbols from the classpath multiple times. We will probably eventually have other contextual data that needs to be shared between transforms, though.

continuationsPlugin/src/test/scala/continuations/ContinuationsPluginSuite.scala
```diff
@@ -72,4 +72,414 @@ class ContinuationsPluginSuite extends FunSuite, CompilerFixtures {
...
+
+      val source =
+        """|
+           |package continuations
+           |
+           |def foo()(using Suspend): Int = {
+           |  val x = 5
+           |  println("HI")
+           |  Continuation.suspendContinuation[Int] { continuation => continuation.resume(Right(1)) }
+           |}
```
jackcviers, line 317: We should probably add this to the coroutines-test to see what it outputs in this case. Can you make a clickup for that and add it as a subtask of the compiler plugin task @gregorath?

continuationsPlugin/src/test/scala/continuations/ContinuationsPluginSuite.scala
```diff
@@ -72,4 +72,414 @@ class ContinuationsPluginSuite extends FunSuite, CompilerFixtures {
...
+
+      val source =
+        """|
+           |package continuations
+           |
+           |def foo()(using Suspend): Int = {
+           |  val x = 5
+           |  println("HI")
+           |  Continuation.suspendContinuation[Int] { continuation => continuation.resume(Right(1)) }
+           |  10
```
jackcviers, line 368: I think this is sane for now. We should probably add this to the coroutines-test to see what it outputs in this case. Can you make a clickup for this one as well, @gregorath?

jackcviers (changes requested):

continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
gregorath: okay moved the vars inside `transformSuspendNoParametersOneContinuationResume` for now in [0cfb287](https://github.com/47deg/TBD/pull/52/commits/0cfb287f3eed96ca5558eda6d6555533e68838d1). 
as you said if we need them elsewhere we can move them to `transformSuspendContinuation` or higher

>Sharing mutable state within the phase

nice, good to know

continuationsPlugin/src/test/scala/continuations/ContinuationsPluginSuite.scala
gregorath: yes I tested that and it keep the rows. 
I will add a task though because it is actually about what it is being returned eg
```
return 10
Continuation.suspendContinuation[Int] { continuation => continuation.resume(Right(1)) }
```

creates a more complicate tree with `switch`.

jackcviers (approved):
Thank you so much for this @gregorath . This PR helped me understand dotty much more deeply. Given what's to come, I think we're gtg at this stage.

PR #53, opened 2022-11-15
30m3gh3 non suspending 0 arity

jackcviers:


continuationsPlugin/src/main/scala/continuations/Continuation.scala
```diff
@@ -13,6 +13,12 @@ object Continuation:
   enum State:
     case Suspended, Undecided, Resumed
 
+  /**
+   * No body needed as the compiler plugin replaces this in call sites
+   */
+  inline def suspendContinuation[A](f: Continuation[A] => Unit)(using s: Suspend): A =
+    ???
```
raulraja, line 20: should probably throw a proper exception unless the plugin synthetically generates this function. If it lives in the classpath compiled and we replace just calls to it, a user may link against the runtime lib without enabling the compiler plugin.

continuationsPlugin/src/main/scala/continuations/DefDefTransforms.scala
```diff
@@ -0,0 +1,65 @@
...
+            if (i == 0) {
+              newPc.appended(completion).asInstanceOf[tpd.ParamClause]
+            } else newPc
+          }.filterNot(_.isEmpty)
+        cpy.DefDef(tree)(paramss = params, tpt = tpd.TypeTree(ctx.definitions.ObjectClass.typeRef))
+    }
+
+  def transformSuspendedDef(tree: tpd.DefDef, cpy: TypedTreeCopier)(
+      using ctx: Context): tpd.DefDef = tree match {
+    case BodyHasNoSuspensionPoint(_) => cpy.DefDef(tree)()
```
raulraja, line 61: Do we still generate the Continuation arg if a suspended function has no suspension points?

raulraja (approved):
Looking great! I may need a small demo soon to understand where we are at

continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
```diff
@@ -1,34 +1,48 @@
...
   import tpd.*
 
-  val phaseName = "continuations"
+  override def phaseName: String = "continuations"
+  override def changesBaseTypes: Boolean = true
+  override def changesMembers: Boolean = true
 
   override val runsAfter = Set(Staging.name)
   override val runsBefore = Set(PickleQuotes.name)
+  override def prepareForDefDef(tree: tpd.DefDef)(using Context): Context = summon[Context]
```
gregorath, line 32: not sure if you have seen it but there is the `dotty.tools.dotc.core.Contexts.ctx` inlined method that does the same (more or less)

continuationsPlugin/src/main/scala/continuations/BodyHasNoSuspensionPoint.scala
```diff
@@ -0,0 +1,32 @@
...
+   *   The method to match upon
+   * @return
+   *   A [scala.Some] if the method has a using [continuations.Suspend] parameter and has a
+   *   suspended continuation in its method body, [scala.None] otherwise
+   */
+  def unapply(defdef: tpd.DefDef)(using Context): Option[tpd.DefDef] =
+    val suspendType = Symbols.requiredMethod("continuations.Continuation.suspendContinuation")
+    Option(defdef).filter {
+      case HasSuspendParameter(_) =>
+        defdef.rhs.existsSubTree{ subtree =>
```
gregorath, line 27: depending on whether we have `suspendContinuation` inlined or not this may not work.

if it is inlined I think the subtrees will traverse the code that is inlined which will be`???` instead of the to-be-inlined one, hence will need something like [this](https://github.com/47deg/TBD/pull/52/files#diff-fade65ebf959155389570d62d274801c2f7607f965a273661e5f776636d6393fR164)

continuationsPlugin/src/test/scala/continuations/CompilerFixtures.scala
```diff
@@ -174,25 +194,298 @@ trait CompilerFixtures { self: FunSuite =>
...
+      )
+
+      val rightOne =
+        ref(right)
+          .select(Names.termName("apply"))
+          .appliedToTypes(List(c.definitions.ThrowableType, intType))
+          .appliedTo(Literal(Constant(1)))
+
+      val rhs =
+        Apply(
```
gregorath, line 376: just a minor, need to decide if we want `inlined` or not because the fixture here (needs to be in `Inlined(...)`) doesn't match the definition [here](https://github.com/47deg/TBD/pull/53/files#diff-120e125acd11ba8a87fb56283fc506e3739f2e28a9bac11688fbd0d0fc4226d9R19) so could be deceiving if try to use it

continuationsPlugin/src/test/scala/continuations/CompilerFixtures.scala
```diff
@@ -174,25 +194,298 @@ trait CompilerFixtures { self: FunSuite =>
...
+    compilerContextWithContinuationsPlugin,
+    zeroArityContextFunctionDefDef
+  )
+
+  val continuationsContextAndSuspendingContextFunctionValDef = FunFixture.map2(
+    compilerContextWithContinuationsPlugin,
+    suspendingContextFunctionValDef
+  )
+
+  val continuationsContextAndNonSuspendingContextFunctionValDef = FunFixture.map2(
```
gregorath, line 475: nice, these fixtures make the code nicer and easier to read for now. and definitely easier to write more tests as transformations are being added

continuationsPlugin/src/test/scala/continuations/CompilerFixtures.scala
```diff
@@ -174,25 +194,298 @@ trait CompilerFixtures { self: FunSuite =>
...
+                Block(
+                  List(DefDef(
+                    anonFunc,
+                    List(List(continuationVal)),
+                    c.definitions.UnitType,
+                    Block(
+                      Nil,
+                      ref(continuationVal).select(Names.termName("resume")).appliedTo(rightOne))
+                  )),
+                  Closure(Nil, ref(anonFunc), TypeTree(c.definitions.UnitType))
```
gregorath, line 393: btw there is another [Closure](https://github.com/lampepfl/dotty/blob/main/compiler/src/dotty/tools/dotc/ast/tpd.scala#L115) which is supposed to create the whole [DefDef Block](https://github.com/47deg/TBD/pull/53/files#diff-d31a1b05e1eddc0b90913369d30ab6c242a20e2ee437a83f0f338652df2b6122R384) but I didn't manage to make it work with that one.

maybe you can give it a go, if it makes the code simpler

continuationsPlugin/src/main/scala/continuations/Continuation.scala
jackcviers: Yeah, I think this should probably be a `Synthetic`. The question is... where do we initialize it? Probably in a prep miniphase `prepareForUnit` -- we only need to do it one time. Ditto for anything we have marked as `???` that we are going to generate. We only need it for typechecking, yeah?

continuationsPlugin/src/main/scala/continuations/DefDefTransforms.scala
jackcviers: The kotlin `coroutines-test` does, and I think we need to if we expect to call `run` at the end of the world.

continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
jackcviers: I actually don't think I need to do this here. We aren't adding anything to the context. It's vestigal from debugging. Removing.

continuationsPlugin/src/test/scala/continuations/CompilerFixtures.scala
jackcviers: It's strange that it works in the compiler plugin suite. I think these are working because I used them in the detection suites, where it may not care if the continuation is inlined or not, since it is using the symbol name.

continuationsPlugin/src/test/scala/continuations/CompilerFixtures.scala
jackcviers: I think it would. I'll give it a go.

continuationsPlugin/src/test/scala/continuations/CompilerFixtures.scala
jackcviers: Agree, we need to add a test case here, good catch @gregorath

continuationsPlugin/src/main/scala/continuations/BodyHasNoSuspensionPoint.scala
jackcviers: Good catch. We need to add a test case using `Inlined`, certainly.

I think the `existsSubTree` uses the `foldover` just like `filterSubTree`:
 
```scala
    def foreachSubTree(f: Tree => Unit)(using Context): Unit = {
      val traverser = new TreeTraverser {
        def traverse(tree: Tree)(using Context) = foldOver(f(tree), tree)
      }
      traverser.traverse(tree)
    }


    /** Is there a subtree of this tree that satisfies predicate `p`? */
    def existsSubTree(p: Tree => Boolean)(using Context): Boolean = {
      val acc = new TreeAccumulator[Boolean] {
        def apply(x: Boolean, t: Tree)(using Context) = x || p(t) || foldOver(x, t)
      }
      acc(false, tree)
    }


    /** All subtrees of this tree that satisfy predicate `p`. */
    def filterSubTrees(f: Tree => Boolean)(using Context): List[Tree] = {
      val buf = mutable.ListBuffer[Tree]()
      foreachSubTree { tree => if (f(tree)) buf += tree }
      buf.toList
    }
```
[Source](https://github.com/lampepfl/dotty/blob/63344e7337727f886d3f7f1a0ad0234dd7670df8/compiler/src/dotty/tools/dotc/ast/tpd.scala#L1102-L1122)

so I don't think there's anything particular about the `Inlined` tree for which the two techniques wouldn't work essentially the same.

They both delegate to `TreeAccumulator` and traverse into each subtree of each tree one at a time in `foldOver`, recursively so should eventually hit the deconstructed `Inlined[T]#call` tree [here](https://github.com/lampepfl/dotty/blob/d8a6751a8d10c73e5a09245cf1456fce4861763b/tests/pos-with-compiler-cc/dotc/ast/Trees.scala#L1583-L1584) and do the same thing as our `filterSubTrees` predicate linked above. We absolutely do need to test it though, because I could be reading the source incorrectly.

continuationsPlugin/src/test/scala/continuations/CompilerFixtures.scala
jackcviers: 😆 I think we may need a better naming scheme. The arity + plugin + context gets kind of wordy. Ideas?

continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
jackcviers: Resolved in https://github.com/47deg/TBD/pull/53/commits/a4ccd1d7a19ecfd29c3d2e627183f9a77957d671

continuationsPlugin/src/main/scala/continuations/BodyHasNoSuspensionPoint.scala
jackcviers: @gregorath -- you were correct. Addressed in https://github.com/47deg/TBD/pull/53/commits/a4ccd1d7a19ecfd29c3d2e627183f9a77957d671

Please resolve the conversation if you agree the problem is addressed.

continuationsPlugin/src/test/scala/continuations/CompilerFixtures.scala
jackcviers: @gregorath Addressed in https://github.com/47deg/TBD/pull/53/commits/a4ccd1d7a19ecfd29c3d2e627183f9a77957d671

Please mark the conversation as resolved if you agree.

continuationsPlugin/src/test/scala/continuations/CompilerFixtures.scala
jackcviers: Yeah, couldn't get the tree to match the compiler output for the tree -- it misses the continuation named argument. Too bad, because it does simplify the code a bit.

continuationsPlugin/src/main/scala/continuations/BodyHasNoSuspensionPoint.scala
gregorath: yeah it is a bit annoying for our case but the `TreeAccumulator` works with the result code of the Inlined. which makes sense I guess, I did try to see if there was a way to say "for now use only the initial code" or if depending on the phase we could get different behaviour but no luck

continuationsPlugin/src/test/scala/continuations/CompilerFixtures.scala
gregorath: ha yes I didn't say anything about that because of the absence of ideas. maybe something like `nonSuspendingVal` and having it in a `object ContinuationFixtures{...}` would also make sense?

naming is hard...

continuationsPlugin/src/main/scala/continuations/DefDefTransforms.scala
```diff
@@ -128,69 +142,76 @@ class DefDefTransforms(using Context) {
...
+  def transformSuspendedDef(tree: tpd.DefDef, cpy: TypedTreeCopier)(
+      using ctx: Context): tpd.DefDef =
     tree match
-      case HasSuspendParameter(_) =>
-        tree match
-          case CallsContinuationResumeWith(resumeArg) =>
-            transformSuspendNoParametersOneContinuationResume(tree, resumeArg)
-          case _ => tree
+      case BodyHasSuspensionPoint(_) => cpy.DefDef(tree)()
+      case HasSuspendParameter(_) => transformNonSuspending(tree, cpy)
```
pabloconde, line 216: This path has already been checked in `transformSuspendContinuation`.

pabloconde (commented):
I don't see that `IsSuspendedType`, `IsCallToSuspend` and `HasInnerSuspensionPoint` classes are being used here, should we remove them?

continuationsPlugin/src/main/scala/continuations/DefDefTransforms.scala
jackcviers: Addressed in https://github.com/47deg/TBD/pull/53/commits/3cbb21aaf70208a1db408fa9e4cb5eda55952e34

jackcviers: > I don't see that IsSuspendedType, IsCallToSuspend and HasInnerSuspensionPoint classes are being used here, should we remove them?

Missed this. I'll clear them out in the morning.

continuationsPlugin/src/main/scala/continuations/Continuation.scala
jackcviers: New card created to address this: https://app.clickup.com/t/371zuqx

PR #55, opened 2022-11-29
CU-43w1hxd - 0-arity suspended after non-suspending exprs and unreachable suspended

pabloconde:
This PR ignores unreachable suspended continuations like:
```
def foo(): Suspend ?=> Int =
  val x = 3
  println("Hi")
  return x
  Continuation.suspendContinuation[Int] { continuation => continuation.resume(Right(1)) }
```

### DIFF
--- a/continuationsPlugin/src/main/scala/continuations/CallsContinuationResumeWith.scala
+++ b/continuationsPlugin/src/main/scala/continuations/CallsContinuationResumeWith.scala
@@ -26,7 +26,14 @@ private[continuations] object CallsContinuationResumeWith:
         .filterSubTrees {
           case Inlined(fun, _, _) =>
             fun.denot.matches(requiredMethod(suspendContinuationFullName))
+          case Return(_, _) => true
           case _ => false
+        }
+        .takeWhile {
+          case Return(Inlined(fun, _, _), _) =>
+            fun.denot.matches(requiredMethod(suspendContinuationFullName))
+          case Return(_, _) => false
+          case _ => true
         }
         .flatMap {
           case Inlined(
--- a/continuationsPlugin/src/test/scala/continuations/CallsContinuationResumeWithSuite.scala
+++ b/continuationsPlugin/src/test/scala/continuations/CallsContinuationResumeWithSuite.scala
@@ -1,9 +1,9 @@
 package continuations
 
 import dotty.tools.dotc.ast.tpd
+import dotty.tools.dotc.ast.tpd.*
 import dotty.tools.dotc.core.Constants.Constant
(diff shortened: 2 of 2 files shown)


continuationsPlugin/src/main/scala/continuations/CallsContinuationResumeWith.scala
```diff
@@ -26,8 +26,15 @@ private[continuations] object CallsContinuationResumeWith:
...
             fun.denot.matches(requiredMethod(suspendContinuationFullName))
+          case Return(_, _) => true
           case _ => false
         }
+        .takeWhile {
+          case Return(Inlined(fun, _, _), _) =>
+            fun.denot.matches(requiredMethod(suspendContinuationFullName))
+          case Return(_, _) => false
+          case _ => true
+        }
```
jackcviers, line 37: Hmm. Wouldn't this fail on something like:

```scala
def mySuspend(x: String)(using Suspend): String =
  if(x == "hello") {
    return x
  } else {
    Continuation.suspend[String](c => c.resumeWith(Right("hello")))
  }
```
where the else is reachable, but there is an early return?

jackcviers (changes requested):

continuationsPlugin/src/main/scala/continuations/CallsContinuationResumeWith.scala
pabloconde: Correct, I'm aware. It would also fail with any similar construct like:

```scala
def mySuspend(x: Int)(using Suspend): Int =
  while (x < 3) {
    return 47
  }
  Continuation.suspendContinuation[Int] { continuation =>
    continuation.resume(Right(1))
  }
```

I need to check if the `return` occurs at the outermost level of the function and filter only that case. I'm still figuring out this.

continuationsPlugin/src/test/scala/continuations/CallsContinuationResumeWithSuite.scala
```diff
@@ -95,3 +88,96 @@ class CallsContinuationResumeWithSuite extends FunSuite, CompilerFixtures:
...
+
+          val foo = defDefs.head.asInstanceOf[tpd.DefDef]
+
+          assertNoDiff(
+            CallsContinuationResumeWith.unapply(foo).get.show,
+            "Right.apply[Nothing, Int](1)")
+      }
+  }
+
+  compilerContext.test("It should return None is the continuation isn't reachable") {
```
gregorath, line 123: hey @pabloconde I can have a look at this scenario as part of the other [story](https://app.clickup.com/t/371z8ck),
unless you have already looked at it

PR #58, opened 2022-12-01
CU-371zuqx - Move suspendContinuation under Suspend and throw an exception

gregorath:
make `suspendContinuation` an extension method of `Suspend` which throws a `CompilerRewriteUnsuccsessfulException` if it has not been transformed by the plugin


continuationsPlugin/src/main/scala/continuations/constants.scala
```diff
@@ -1,8 +1,15 @@
...
+
 private[continuations] val continuationPackageName = "continuations"
+private[continuations] val suspendContinuationName = "suspendContinuation"
 private[continuations] val suspendFullName = s"$continuationPackageName.Suspend"
 private[continuations] val continuationFullName = s"$continuationPackageName.Continuation"
-private[continuations] val suspendContinuationFullName =
-  s"$continuationFullName.suspendContinuation"
 private[continuations] val resumeFullName = s"$continuationPackageName.Continuation.resume"
+
+private[continuations] def suspendContinuationMethod(using Context): Select =
```
gregorath, line 14: not a constant per se but not sure where else it would make sense to have it

continuationsPlugin/src/main/scala/continuations/Suspend.scala
```diff
@@ -1,9 +1,22 @@
 package continuations
 
+import continuations.Suspend.CompilerRewriteUnsuccessfulException
+
 sealed trait Suspend:
   def continuation[A](f: Continuation[A] => Unit): A
 
+  extension (s: Suspend)
```
gregorath, line 8: is that what you had in mind @jackcviers ?

continuationsPlugin/src/main/scala/continuations/Suspend.scala
```diff
@@ -1,9 +1,22 @@
...
+    inline def suspendContinuation[A](f: Continuation[A] => Unit): A =
+      throw CompilerRewriteUnsuccessfulException
+
 object Suspend:
+  case object CompilerRewriteUnsuccessfulException
+      extends Exception(
+        """|
+           |  The body of this method should have been generated by the continuations-plugin.
+           |  Please open an issue so we can fix this:
+           |  https://github.com/47deg/TBD/issues/new?assignees=&labels=bug&template=compilerrewriteunsuccessfulexception-bug-report.md&title=CompilerRewriteUnsuccessfulException+Bug""".stripMargin)
```
gregorath, line 18: the link probably needs update but it will point to this one https://github.com/47deg/TBD/pull/56

jackcviers: I approved #56. We should merge that and update the link.

continuationsPlugin/src/main/scala/continuations/Suspend.scala
```diff
@@ -1,9 +1,22 @@
...
+
 sealed trait Suspend:
   def continuation[A](f: Continuation[A] => Unit): A
 
+  extension (s: Suspend)
+    inline def suspendContinuation[A](f: Continuation[A] => Unit): A =
+      throw CompilerRewriteUnsuccessfulException
+
 object Suspend:
+  case object CompilerRewriteUnsuccessfulException
```
jackcviers, line 13:
```suggestion
  private[continuations] case object CompilerRewriteUnsuccessfulException
```
We don't want anyone else calling this.

continuationsPlugin/src/main/scala/continuations/constants.scala
jackcviers: Let's create a `Trees` trait and put it under that, and mix it in where we need it. We could probably extract a lot of common Tree's there, eventually.

continuationsPlugin/src/test/scala/continuations/ContinuationsPluginSuite.scala
```diff
@@ -429,6 +429,52 @@ class ContinuationsPluginSuite extends FunSuite, CompilerFixtures {
       }
   }
 
+  compilerContextWithContinuationsPlugin.test(
+    "it should convert simple suspended def with a named implicit Suspend") {
+    implicit givenContext =>
```
jackcviers, line 434:
```suggestion
    case given Context =>
```

jackcviers (changes requested):

continuationsPlugin/src/main/scala/continuations/Suspend.scala
gregorath: ha it actually looks like the link is correct. I hand wrote it so I was expecting to have missed something. but apparently it is good.

continuationsPlugin/src/test/scala/continuations/ContinuationsPluginSuite.scala
gregorath: done in [4b8f514](https://github.com/47deg/TBD/pull/58/commits/4b8f51434dce80fdbe7b7d9a776119e8d333bf68)

continuationsPlugin/src/main/scala/continuations/constants.scala
gregorath: done in [f01e32f](https://github.com/47deg/TBD/pull/58/commits/f01e32f3053eabccc7eedc642106aced40394ebb)

jackcviers (approved):

PR #61, opened 2022-12-15
Report error if there is a nested continuation

pabloconde:


continuationsPlugin/src/main/scala/continuations/CallsContinuationResumeWith.scala
```diff
@@ -24,8 +32,13 @@ private[continuations] object CallsContinuationResumeWith extends Trees:
       tree
         .rhs
         .filterSubTrees {
-          case Inlined(fun, _, _) =>
-            fun.denot.matches(suspendContinuationMethod.symbol)
+          case Inlined(call, _, _) =>
+            val isSuspendContinuation = call.denot.matches(suspendContinuationMethod.symbol)
+            if isSuspendContinuation && hasNestedContinuation(call) then
+              report.error(
```
gregorath, line 38: long shot but I guess there is no way to collect those errors for the tests?

gregorath (approved):
lgtm.

continuationsPlugin/src/main/scala/continuations/CallsContinuationResumeWith.scala
```diff
@@ -4,13 +4,21 @@ import dotty.tools.dotc.ast.tpd.*
...
 import dotty.tools.dotc.core.Names.*
+import dotty.tools.dotc.*
 
 /**
  * Matcher for detecting methods that call and [[continuations.Suspend#suspendContinuation]]
  * [[continuations.Continuation.resume
  */
 private[continuations] object CallsContinuationResumeWith extends Trees:
 
+  private def hasNestedContinuation(tree: Tree)(using Context): Boolean =
```
gregorath, line 15: ah could this maybe be replaces by [this](https://github.com/47deg/TBD/blob/compiler-plugin/continuationsPlugin/src/main/scala/continuations/TreesChecks.scala#L9)?

pabloconde: Unfortunately no, if the subtree is not `Inlined` I want it to be `false` by default; the other function would always match the caller tree, regardless of wether there is a nested continuation.

I know we have many functions that do similar things throughout the code, but I preferred to wait for the code mentioned [here](https://github.com/47deg/TBD/pull/60#discussion_r1045976303) to be merged.

gregorath: okay yes.
I am not sure how the default case will work in the other function, because `suspendContinuation` will always be inlined so probably it will always be false?  but anyway you are right we can do some cleaning in the end

continuationsPlugin/src/main/scala/continuations/CallsContinuationResumeWith.scala
jackcviers: I haven't figured out how to do it yet. It's probably in the compiler tests in the dotty repo if you want to look.

PR #63, opened 2023-01-12
CU-43w1jzc - Compiler Plugin: Transform zero arity method with unused suspend continuation that returns non-suspend value

gregorath:
this re-writes and works for the following code
```
def programSuspendContinuationNoParamResumeIgnoreResult: Int =
  def fooTest()(using s: Suspend): Int =
    s.suspendContinuation[Unit] { _.resume(Right { println("Hello") }) }
    10

  fooTest()
```
~I have copied some code from [Jack's branch](https://github.com/jackcviers/TBD/tree/30m3gt1-single-arity-suspend-with-non-suspend-dependent-return-solutions) eg: `prepareForUnit`, `prepareForDefDef`, `createSyntheticNames`, `countContinuationSynthetics` etc.~ removed at [6029efb](https://github.com/47deg/TBD/pull/63/commits/6029efb2becc9d4aed337cee7e38127bd9044015)

I am not sure if we want to merge this code as is or use it as reference and move bits/combine with Jack's work (where we use the result of `suspendContinuation`).

However for now it doesn't work if:
-  ~there are more than one `suspendContinuation` in the body. So it doesn't create the whole state machine but only the most basic one (I will look at this next)~ This is now supported after [6029efb](https://github.com/47deg/TBD/pull/63/commits/6029efb2becc9d4aed337cee7e38127bd9044015) as well as the case beneath
```
def programSuspendContinuationNoParamResumeIgnoreResult: Int =
  def fooTest()(using s: Suspend): Int =
    s.suspendContinuation[Unit] { _.resume(Right { println("Hello") }) }
    s.suspendContinuation[Unit] { _.resume(Right { println("Hello") }) } // *1,2,3,...
    10

  fooTest()
 ``` 

```
def programSuspendContinuationNoParamResume: Int =
  def fooTest()(using s: Suspend): Unit =
    s.suspendContinuation[Unit] { continuation => continuation.resume(Right { println("Hi") }) }  // *1,2,3,...
    s.suspendContinuation[Int] { continuation => continuation.resume(Right(1)) }

  fooTest()
```
- `fooTest` is inside an `object` or `class`. I noticed that this is true for the other transformations we have as well so we need to look at it probably separately as we only worked to move the Kotlin code for this exact scenario we show here.
- ~there are [more than 1 resume ](https://github.com/gregorath/TBD/blob/9c285cd34254e52ca199169b2d8f3ff0b0cba6ce/continuationsPlugin/src/main/scala/continuations/DefDefTransforms.scala#L577)~  fixed in 74ddd74 + 
~our method detection (`CallsContinuationResumeWith.scala`) at the moment will return None if there are more than 1 resume~(changed in [6029efb](https://github.com/47deg/TBD/pull/63/commits/6029efb2becc9d4aed337cee7e38127bd9044015))
- ~the [detection points](https://github.com/47deg/TBD/blob/9c285cd34254e52ca199169b2d8f3ff0b0cba6ce/continuationsPlugin/src/main/scala/continuations/HasSuspensionNotInReturnedValue.scala) are also a bit basic for now, eg if we have a `return suspendContinuation(...)` at some place inside the `Block`~ improved in [8ce5f91](https://github.com/47deg/TBD/pull/63/commits/8ce5f91a0ac5f5cb035d31004a694b3e31127c30)

- ~there is code between the suspensions:~ done in [56dd49a](https://github.com/47deg/TBD/pull/63/commits/56dd49abe8fc80d3c9036958e55a95643832f8e9)
```
def programSuspendContinuationNoParamResumeIgnoreResult: Int =
  def fooTest()(using s: Suspend): Int =
    val x = 1
    s.suspendContinuation[Unit] { _.resume(Right { println("Hello") }) }
    println(s"Hello $x")
    s.suspendContinuation[Unit] { _.resume(Right { println("Hello") }) } // *1,2,3,...
    10

  fooTest()
 ```


gregorath:
@jackcviers @pabloconde this has been updated now to support multiple suspensions, have a look at the initial description about what doesn't work (yet).
the code could be cleared a bit but I guess this is something to look at once we have everything and we combine the scenarios.

I follow kind of a different way from Jack although I have added `Labelled + Return` too but I think it could be a good guide for doing the dependent suspensions as well

gregorath:
hey @jackcviers  @pabloconde  now the PR is ready for final review. 
It works with one or more `suspendContinuation` which are not assigned to a val, building the whole state machine etc.
I added a [stub continuation](https://github.com/47deg/TBD/pull/63/files#diff-8a3ec0f102df6f960a95faa4e9caf4241005cde85491000c9e24ab0f59aefb44R5) to be able to run the examples.

on top of that I fixed the existing examples we had to work with the `stub continuation` as they were broken and also I added support for context functions, multiple `resume` clauses, keeping any other code around the `suspendContinuation` or the `resume`.
For the existing cases I changed the params to work with type params as well but this is maybe something we want to improve later on as for now it merges all the param groups to one, see [here](https://github.com/47deg/TBD/pull/63/commits/1683a4969cd66d040a11b333f693a9fdb5d5f08a#diff-24d2311d1a38e4cfc4a1e3df45fd867f82ca2116403a8b5c0b45ecf07fbc2223R2053) (but for now it run as expected)

continuationsPlugin/src/main/scala/continuations/ContinuationsPlugin.scala
```diff
@@ -1,32 +1,114 @@
...
-import dotty.tools.dotc.core.Contexts.Context
+import dotty.tools.dotc.core.Contexts.{ctx, Context}
+import dotty.tools.dotc.core.Flags
+import dotty.tools.dotc.core.Symbols.{
+  requiredClass,
+  requiredClassRef,
+  requiredMethod,
+  requiredModule,
+  Symbol
+}
```
pabloconde, line 12: Nit: maybe we could just write `import dotty.tools.dotc.core.Symbols.*`

pabloconde (approved):
LGTM. Have seen it working 🎉

jackcviers (approved):
Looks great, and runs great. Awesome work @gregorath

PR #64, opened 2023-01-31
CU-865bm99yr - Compiler Plugin: Add examples for demo

gregorath:
this is on top of https://github.com/47deg/TBD/pull/63, look at commit c43ea88.
just added a few examples to see how it would look like, we will add more closer to the demo day


jackcviers:
Is this ready for review @gregorath

PR #67, opened 2023-02-06
CU-865bm99yr - Compiler Plugin: Add examples for demo

gregorath:
just added a few examples to see how it would look like, we will add more closer to the demo day once we have agreed on what we are presenting


continuationsPluginExample/src/main/scala/examples/ListMap.scala
```diff
@@ -0,0 +1,12 @@
...
+
+import continuations.Suspend
+
+@main def ListMap =
+  def twoArgumentsOneContinuationsCFBefore(x: Int, y: Int): Suspend ?=> Int =
+    val z = 1
+    summon[Suspend].suspendContinuation[Int] { continuation =>
+      continuation.resume(Right(x + y + z))
+    }
+  val mappedContinuations = List(1, 2, 3, 4).map(twoArgumentsOneContinuationsCFBefore(1, _))
```
gregorath, line 11: 👏

notes/TwoArgumentsTwoContinuationsDecompiled.txt
```diff
@@ -0,0 +1,184 @@
+/*
+ * Decompiled with CFR 0.152.
```
gregorath, line 2: I have to say in case people want to look more at the logic, even though this is closer to how our code looks like, I found the outcome from `Procyon` decompiler easier to read and understand.

gregorath (commented):
nice, splitting these out makes it easier to follow. looking forward to the demo.

I will add the names of the ones that [were not kept ](https://github.com/47deg/TBD/blob/c14bd71fbb8a26e711e91fb71a7f16fadee54f37/continuationsPluginExample/src/main/scala/examples/DemoExamples.scala) in case we want to reference them from the spreadsheet 
```
twoCurriedArgumentsZeroContinuations
oneArgumentsSingleResumeContinuationsBeforeAfter
twoArgumentsSingleResumeContinuationsBeforeAfter
useValsDefinedInsideResume
```

It looks like I can't approve it but 👍

jackcviers:
> It looks like I can't approve it

You can just comment "Approved" and I'll merge with Admin privileges.

pabloconde (approved):

PR #70, opened 2023-02-07
CU-865bm99yr - Compiler plugin: Fix calling a suspending function with generic params

gregorath:
while looking at the scenarios for the demo we found with @pabloconde that this case was not working (it was working only for A == Int, String, etc). opening a separate fix for it  instead of having it as part of the demo PR


pabloconde (approved):
LGTM

PR #71, opened 2023-02-08
CU-43w1jzc - Compiler Plugin: Set ownership if there are vals only in resume body

gregorath:
Looking at the demo cases I found that one was not working because the vals' ownership was not being updated. since it is a small fix I open a PR on the story that this functionality was implemented initially


pabloconde (approved):
Good catch

jackcviers (approved):

PR #73, opened 2023-02-10
8677bvbh0: Creates munit-snap shapshot testing utility.

jackcviers:


jackcviers:
https://app.clickup.com/t/8677bvbh0

munit-snap/src/main/scala/munit/SnapshotSuite.scala
```diff
@@ -0,0 +1,130 @@
...
+
+  given codec[A: Encoder: Decoder]: Codec[A] =
+    ic.Codec.from(summon, summon)
+
+  /**
+   * Given a json codec for the type of the result of the snapshot test body, execute a snapshot
+   * test on the body results.
+   *
+   * The first time the test is run, the json encoded result is recorded to a file in the same
+   * directory as the test file using the location, test name, and line number of the snapshot
```
gregorath, line 42: could we maybe have a directory `snapshot-files` or something where all the json files with the snapshotted values will exist? 
otherwise with the number of tests we have I think it will be a bit hard to navigate in the tests folder trying to find the `...Suite` ones.

gregorath:
really nice work and valuable I am looking forward to this. just a small comment to discuss

munit-snap/src/main/scala/munit/SnapshotSuite.scala
jackcviers: Sure. I'd also really like a Tag to allow for resetting a test. If you use .reset (like .only or .debug) the snapshot is deleted and the test passes and stores the new result.

jackcviers: It might also be better to also store the hashes of the result. Compare them first before reading the file. Some files could get quite large.

pabloconde: That tag sounds very useful. I'm looking forward to this work!

gregorath: >If you use .reset (like .only or .debug) the snapshot is deleted and the test passes and stores the new result.
that would be really nice yes. either in this PR or later I guess

gregorath (approved):
🎉

PR #74, opened 2023-02-14
[Compiler Plugin] - CU#8677bhj55 - Dependent suspend continuations with two arguments

pabloconde:
This PR adds some missing test cases for dependent continuations with two arguments.

### DIFF
--- a/continuationsPluginExample/src/main/scala/examples/Main.scala
+++ b/continuationsPluginExample/src/main/scala/examples/Main.scala
@@ -12,6 +12,7 @@ import continuations.jvm.internal.ContinuationImpl
   println(result3)
   val result4 = programSuspendContinuationNoParamResumeIgnoreResult
   println(result4)
+  println(programSuspendContinuationParamDependent)
 
 object main$handler extends (Continuation[Any | Null] => Any):
   override def apply($completion: Continuation[Any | Null]): Any =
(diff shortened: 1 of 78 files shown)


continuationsPluginExample/src/main/scala/examples/Main.scala
```diff
@@ -6,13 +6,8 @@ import continuations.jvm.internal.ContinuationImpl
...
 //  println(result)
-  val result2 = programSuspendContinuationNoParamResume
-  println(result2)
-  val result3 = programSuspendContinuationNoParamNoSuspendContinuation
-  println(result3)
-  val result4 = programSuspendContinuationNoParamResumeIgnoreResult
-  println(result4)
-  println(programSuspendContinuationParamDependent)
+
+  println(programSuspendContinuationTwoParamsDependent)
```
gregorath, line 10: should we not keep the existing ones too?

jackcviers: I think we should.

jackcviers: Also: there should be a standalone main for `programSuspendContinuationTwoParamsDependent` like we have for the others we just demoed.

jackcviers (changes requested):

gregorath (approved):
👍

jackcviers (approved):

PR #75, opened 2023-02-17
CU-8677bkvqn - Compiler Plugin: ValDef with NonSuspending

gregorath:

PR #76, opened 2023-02-21
Adds snapshot example

jackcviers:


continuationsPlugin/src/it/scala/continuations/CompilerFixtures.scala
```diff
@@ -0,0 +1,896 @@
...
+import munit.FunSuite
+
+import scala.concurrent.Promise
+import scala.util.Properties
+import scala.util.Try
+
+trait CompilerFixtures { self: FunSuite =>
+  def continuationsCompilerSnapshot(source: String)(using Context): (String, String) =
+    val p = Promise[(String, String)]
+    checkContinuations(source)((t, c) => p.complete(Try((t.show, c.toString))))
```
gregorath, line 136: I am not quite sure what having the context here gives us, as you can that the information in the file doesn't show anything useful.
also it includes things like `MutableScope@163a9512` which is an instance that has a different name on every run so actually it can't be used. I tried to use the `c.scope.toList` but that comes back empty

jackcviers:
Closed as it was an example and was integrated with #75

PR #79, opened 2023-02-27
Adds ThreeDependentContinuations example

jackcviers:
From Demo

### DIFF
--- a/continuationsPluginExample/src/main/scala/examples/ThreeDependentContinuations.scala
+++ b/continuationsPluginExample/src/main/scala/examples/ThreeDependentContinuations.scala
@@ -1,0 +1,14 @@
+package examples
+
+import continuations.*
+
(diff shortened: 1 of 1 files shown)


pabloconde (approved):

diesalbla (approved):
Is there a way of testing or showing what the "generated" code is, so we may keep it as a regression for next changes?

gregorath:
in this case we are adding the code as an example for demo purposes. when you compile/run the example you can see the output.

regarding the regression tests we recently adding support for snapshot tests, [here](https://github.com/47deg/TBD/pull/73) is the functionality and [here](https://github.com/47deg/TBD/tree/main/continuationsPlugin/src/it/scala/continuations) adding some snapshot tests. 
the plan is to add more of them for the other cases too but haven't done it yet as the functionality was added recently

PR #81, opened 2023-03-01
CU--8677bmjpu - Compiler Plugin: Suspend in non-companion object

gregorath:
in this PR we added the tests however as discussed there is an issue with sbt, so for now it only works with scalac. there are a couple of issues with the transformation itself but we need sbt in order to debug them as we need to look at our phase

- sbt returns an error:

`Failed to find name hashes for examples.program$.program$foo$1`
where `foo` is the class we create for the state machine ` class program$foo$1($completion: continuations.Continuation[Any | Null])`

- however there seems to be a bug in sbt, see https://github.com/sbt/sbt/issues/ 7157

using `scalac:3.1.2` or `scala` seems to create the correct tree and work

if the code from `NonCompanionObjectContinuations` is copied into `ZeroArgumentsZeroContinuations.scala`
```
env JAVA_OPTS='--enable-preview --add-modules jdk.incubator.concurrent' scalac -classpath continuationsPlugin/target/scala-3.1.2/classes -d zero-arguments-no-continuation-treeview/target/scala-3.1.2/classes -Xplugin:continuationsPlugin/target/scala-3.1.2/continuationsplugin_3-0.0.0+146-d84ba681+20230227-1101-SNAPSHOT.jar zero-arguments-no-continuation-treeview/src/main/scala/ZeroArgumentsZeroContinuations.scala
```


continuationsPlugin/src/it/scala/continuations/NonCompanionObjectSnapshotsSuite.scala
```diff
@@ -0,0 +1,465 @@
...
+          |
+          |    method3(x) + z1 + z3 + 1
+          |  }
+          |}
+          |""".stripMargin
+
+      continuationsCompilerSnapshot(source)
+  }
+
+  compilerContextWithContinuationsPlugin.snapshotTest("test-3".ignore) {
```
gregorath, line 77: tests with `ignore` contain code that fails when it runs so we need to debug them. the test itself passes because the tree seems okay but in running with `scalac` it has an error. unfortunately we need sbt in order to debug these

continuationsPluginExample/src/main/scala/examples/CPSExample.scala
```diff
@@ -215,3 +215,38 @@ def programSuspendContinuationNoSuspendContinuationVal: Int =
                 x.size + y.size + q.size + p.size + z
 
   fooTest(List(1))(List("A", "B"))(List(1, 1, 1))(List(1, 1, 1, 1))
+
+// runs only with scalac
```
jackcviers, line 219: ContinuationsPluginExample should use scalac for compilation now. What happens when you uncomment this section?

continuationsPluginExample/src/main/scala/examples/NonCompanionObjectContinuations.scala
```diff
@@ -0,0 +1,41 @@
+//package examples
```
jackcviers, line 1: Should compile now with the updates to main to delegate to scalac via coursier launch.

continuationsPluginExample/src/main/scala/examples/CPSExample.scala
jackcviers: You can uncomment this section.

continuationsPluginExample/src/main/scala/examples/NonCompanionObjectContinuations.scala
jackcviers: You can uncomment this.

jackcviers: On main, there should be a `bypass-zinc` project that aggregates the projects that need raw scalac compilation, and changes to the ci alias that keep it from crashing zinc in ci: https://github.com/47deg/TBD/blob/56cf64c3027716558597f2c283572a6f4f9f2d5f/build.sbt#L17-L42

continuationsPluginExample/src/main/scala/examples/CPSExample.scala
jackcviers: On main, there should be a `bypass-zinc` project that aggregates the projects that need raw scalac compilation, and changes to the ci alias that keep it from crashing zinc in ci: https://github.com/47deg/TBD/blob/56cf64c3027716558597f2c283572a6f4f9f2d5f/build.sbt#L17-L42

jackcviers (changes requested):

continuationsPluginExample/src/main/scala/examples/NonCompanionObjectContinuations.scala
gregorath: thanks for flagging this.
done in [77b9be8](https://github.com/47deg/TBD/pull/81/commits/77b9be8fdbbd5dfd4ba461589c6b9cae4acc0367)

continuationsPluginExample/src/main/scala/examples/CPSExample.scala
gregorath: done in [77b9be8](https://github.com/47deg/TBD/pull/81/commits/77b9be8fdbbd5dfd4ba461589c6b9cae4acc0367)

diesalbla (approved):

continuationsPlugin/src/it/scala/continuations/NonCompanionObjectSnapshotsSuite.scala
```diff
@@ -0,0 +1,465 @@
+package continuations
+
+import dotty.tools.dotc.core.Contexts.Context
+import munit.SnapshotSuite
+
+class NonCompanionObjectSnapshotsSuite extends SnapshotSuite, CompilerFixtures:
+
+  compilerContextWithContinuationsPlugin.snapshotTest("test-1") {
```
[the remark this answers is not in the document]
gregorath: okay yes it makes sense. I will have a look

jackcviers (approved):

continuationsPlugin/src/it/scala/continuations/NonCompanionObjectSnapshotsSuite.scala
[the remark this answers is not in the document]
gregorath: done in [dfe1cd3](https://github.com/47deg/TBD/pull/81/commits/dfe1cd3a71c73620ccea6723ed628658889473d4)

PR #82, opened 2023-03-01
DefDefTransforms - reading refactorings...

diesalbla:
Some refactorings as I was reading through the code, to extract auxiliary functions (specially predicates for filters).


jackcviers:
Looks like one of the refactorings collapsed a closure block or the anonymous function symbol wasn't fully replaced, or in a tree with _.resume the referred parameter wasn't replaced with safeContinuation.

This should only be possible in params or transformSuspensionPoint.

I'll pull and point out what broke the tests shortly.

diesalbla:
@jackcviers I think I found the cause for the error.

I rewrote a `xss.flatten.filterNot( p ) .isEmpty` into a `xss.forall(_.forall(x => ! p(x))` but it was the negation. I force-pushed correction so tests are passing now.

jackcviers (approved):
Thanks for extracting these!

PR #83, opened 2023-03-02
Introduce a BlockFlattener transformation.

diesalbla:
The generated code sometimes includes blocks within blocks within blocks, and there is no good reason to have that. So we add a BlockFlattener transformation to clean those up.

We define it as a class outside of any method to emphasize that the transformation is well stand-alone, and could very well be part of a general macro utility.

### DIFF
--- a/continuationsPlugin/src/main/scala/continuations/DefDefTransforms.scala
+++ b/continuationsPlugin/src/main/scala/continuations/DefDefTransforms.scala
@@ -28,9 +28,20 @@ object DefDefTransforms extends TreesChecks:
   def flattenBlock(b: tpd.Block)(using Context): tpd.Block =
     b match {
       case bb @ Trees.Block(Nil, _: tpd.Closure) => bb
-      case Trees.Block(Nil, bb @ tpd.Block(_Nil, _)) => bb
+      case Trees.Block(Nil, bb @ tpd.Block(_, _)) => bb
       case b => b
     }
+
+  private class BlockFlattener(using Context) extends TreeMap {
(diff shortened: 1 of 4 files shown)


gregorath:
nice 👍 

but is this a preparation for https://github.com/47deg/TBD/pull/80/files?

and is this something that could be applied in others cases like we see here https://github.com/47deg/TBD/blob/473cc3418f5b8c5b88a38fe139033fbc1f02b7ab/continuationsPlugin/src/test/scala/continuations/StateMachineFixtures.scala#L64? 
pretty much what will happen if we apply it to the final tree we create instead of just the body of `resume?

jackcviers:
We do need to be careful about where we apply this because it can change the meanings of assignments @gregorath , so we have to only apply it where the nested block generation is synthetic. We know that happens in #80 in the transformation of the body.

If there are places where we know we produce these blocks, it's useful to fix our output and we can apply it.

gregorath:
okay yes that makes sense. better to keep a safe approach to it. we did have more unnecessary nested blocks in the past I think but seems like that is not the case anymore, apart from that link

@diesalbla happy to ignore my previous comment :)

jackcviers (changes requested):
Fix the scalafmt.

Did you run it against the specific test cases I passed in the stub example branch @diesalbla?

diesalbla:
> Did you run it against the specific test cases I passed in the stub example branch @diesalbla?
 
Not yet. Let me try

jackcviers (approved):

PR #85, opened 2023-03-02
More refactorings - local vals, add comments

diesalbla:


continuationsPlugin/src/main/scala/continuations/DefDefTransforms.scala
```diff
@@ -25,6 +25,15 @@ import scala.collection.mutable.ListBuffer
 
 object DefDefTransforms extends TreesChecks:
 
+  def transformSuspendContinuation(tree: tpd.ValOrDefDef)(using Context): tpd.Tree =
```
gregorath, line 28: you may want to move `def transformContinuationWithSuspend` below this one as well?

gregorath (approved):
nice these refactorings make the code easier to read

PR #86, opened 2023-03-03
Refactor: turn matchers to predicates.

diesalbla:
If the tree you may return in the `Some` is the same as the input, that is  not a matcher, that is a predicate. However, is there anything else going on in this processes? 

Otherwise, does the

### DIFF
--- a/continuationsPlugin/src/main/scala/continuations/DefDefTransforms.scala
+++ b/continuationsPlugin/src/main/scala/continuations/DefDefTransforms.scala
@@ -73,9 +73,9 @@ object DefDefTransforms extends TreesChecks:
         report.logWith("state machine and new defdef:")(transformedTree)
 
     tree match
-      case HasSuspensionNotInReturnedValue(_) =>
+      case _ if HasSuspensionNotInReturnedValue(tree) =>
         transformSuspensionsSuspendingStateMachine(fetchSuspensions, false)
-      case CallsSuspendContinuation(_) =>
+      case _ if CallsSuspendContinuation(tree) =>
         fetchSuspensions match
           case suspensionPoint :: Nil if !suspensionPoint.isInstanceOf[tpd.ValDef] =>
             transformSuspendOneContinuationResume(tree, suspensionPoint)
--- a/continuationsPlugin/src/main/scala/continuations/DefDefTransforms.scala
+++ b/continuationsPlugin/src/main/scala/continuations/DefDefTransforms.scala
@@ -27,9 +27,10 @@ object DefDefTransforms extends TreesChecks:
 
   def transformSuspendContinuation(tree: tpd.ValOrDefDef)(using Context): tpd.Tree =
     tree match {
-      case ReturnsContextFunctionWithSuspendType(_) if !tree.symbol.isAnonymousFunction =>
+      case _
+          if ReturnsContextFunctionWithSuspendType(tree) && !tree.symbol.isAnonymousFunction =>
         report.logWith(s"new tree:")(transformContinuationWithSuspend(tree))
-      case HasSuspendParameter(_) if !tree.symbol.isAnonymousFunction =>
+      case _ if HasSuspendParameter(tree) && !tree.symbol.isAnonymousFunction =>
         report.logWith(s"new tree:")(transformContinuationWithSuspend(tree))
       case _ => report.logWith(s"oldTree:")(tree)
     }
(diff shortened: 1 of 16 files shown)


gregorath:
I believe this is a pattern we saw in the compiler code and followed it as well. but yeah it is the same one way or another

jackcviers:
The idea was to allow us to use it to extract trees that matched and their subtrees for use in the transformations. It didn't end up being used as I originally hoped in the tree maps. We can certainly change them to be predicates instead.

continuationsPlugin/src/main/scala/continuations/DefDefTransforms.scala
```diff
@@ -72,19 +72,19 @@ object DefDefTransforms extends TreesChecks:
...
             transformSuspendOneContinuationResume(tree, suspensionPoint)
           case suspensionPoints =>
             transformSuspensionsSuspendingStateMachine(suspensionPoints, true)
-      case BodyHasSuspensionPoint(_) =>
+      case vd: tpd.DefDef if BodyHasSuspensionPoint(vd) =>
         // any suspension that still needs a transformation
         tree match
           case t: tpd.DefDef => cpy.DefDef(t)()
-          case t: tpd.ValDef => cpy.ValDef(t)()
+      // case t: tpd.ValDef => cpy.ValDef(t)()
```
gregorath, line 87: this is necessary because tree can be def or val https://github.com/47deg/TBD/blob/23c21f31779182f74ae922ec0cae1a872a92b20f/continuationsPlugin/src/main/scala/continuations/DefDefTransforms.scala#L55.

we did some work on `val` but we de-prioritised for now

continuationsPlugin/src/test/scala/continuations/CallsSuspendContinuationSuite.scala
```diff
@@ -11,34 +11,14 @@ import munit.FunSuite
 
 class CallsSuspendContinuationSuite extends FunSuite, CompilerFixtures:
 
-  continuationsContextAndZeroAritySuspendSuspendingDefDef.test(
```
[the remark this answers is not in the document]
diesalbla: Well, can the matchers alter the tree? Or do we need to make sure that they have _not_ altered it?

gregorath (approved):

continuationsPlugin/src/test/scala/continuations/CallsSuspendParameterSuite.scala
```diff
@@ -12,7 +12,9 @@ class CallsSuspendParameterSuite extends FunSuite, CompilerFixtures {
       val call = tree match
         case Block(_, call) => call
 
-      assertNoDiff(CallsSuspendParameter.unapply(call).get.show, call.show)
+      val prev = call.show
+      val _ = CallsSuspendParameter(call)
+      assertNoDiff(call.show, prev)
```
gregorath, line 17: probably should change this one to check it is `true`

continuationsPlugin/src/test/scala/continuations/HasSuspendParameterSuite.scala
```diff
@@ -8,18 +8,18 @@ class HasSuspendParameterSuite extends FunSuite, CompilerFixtures {
   continuationsContextAndZeroAritySuspendNonSuspendingDefDef.test(
     "#unapply(defDefTree): def mySuspend()(using Suspend): Int = 1 should be Some(tree)") {
```
gregorath, line 9: ideally we would like to change the test descriptions too? 😬

continuationsPlugin/src/test/scala/continuations/CallsSuspendParameterSuite.scala
gregorath: after discussing with @diesalbla it seems that some of these predicates change the tree in a way that affects `.show`.
for example [here](https://github.com/47deg/TBD/blob/main/continuationsPlugin/src/test/scala/continuations/HasSuspensionNotInReturnedValueSuite.scala#L33) if we do the same comparison as above we see this difference
```
=> Diff (- obtained, + expected)
-val mySuspend: ((continuations.Suspend : Object {...})) ?=> Int = ∙
+val mySuspend: (continuations.Suspend.type) ?=> Int = ∙
```

the trees though before and after are the same, if we don't use `.show`. 

since this seems to not affect our transformation and we use these predicates only to see if they return true/false I think this is irrelevant for now and if we want to investigate it more then should do separately

pabloconde (approved):
LGTM

PR #88, opened 2023-03-03
Extract "checkResult" procedure in generated code.

diesalbla:
This reduces the size of the generated code.

_Note_ still fixing tests at time of opening PR.

### DIFF
--- a/continuationsPlugin/src/main/scala/continuations/Continuation.scala
+++ b/continuationsPlugin/src/main/scala/continuations/Continuation.scala
@@ -13,6 +13,15 @@ object Continuation:
   enum State:
     case Suspended, Undecided, Resumed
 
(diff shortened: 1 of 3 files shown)


jackcviers (approved):

continuationsPlugin/src/main/scala/continuations/Continuation.scala
```diff
@@ -13,6 +13,15 @@ object Continuation:
   enum State:
     case Suspended, Undecided, Resumed
 
+  type Result = Either[Throwable, Any | Null | State.Suspended.type]
+
+  def checkResult(result: Result): Unit =
```
gregorath, line 18: nice that was a good idea.

PR #89, opened 2023-03-03
Upgrade SBT and Scala versions

diesalbla:


pabloconde (commented):
Could we update the `.sdkmanrc` file as part of this PR as well?

jackcviers (changes requested):
Looks like the mdoc for this scala version fails in the tests. Can you fix that?

PR #90, opened 2023-03-03
Code Generation - Simplify completion pattern

diesalbla:
We can remove the use of an `isInstanceOf` clause and instead use a typed pattern


continuationsPlugin/src/main/scala/continuations/DefDefTransforms.scala
```diff
@@ -863,32 +863,25 @@ object DefDefTransforms extends TreesChecks:
...
-              .select(labelVarParam)
-              .select(integerAND)
-              .appliedTo(integerMin)
-              .select(integerNE)
-              .appliedTo(tpd.Literal(Constant(0x0)))),
+          .select(labelVarParam)
+          .select(integerAND)
+          .appliedTo(integerMin)
+          .select(integerNE)
+          .appliedTo(tpd.Literal(Constant(0x0))),
```
jackcviers, line 879: 👍🏻

continuationsPlugin/src/test/scala/continuations/StateMachineFixtures.scala
```diff
@@ -1149,22 +1149,16 @@ trait StateMachineFixtures {
...
        |                      compileFromString$package$foo$1
        |                  ].$label.&(scala.Int.MinValue).!=(0)
-       |                )
        |               =>
-       |                $continuation =
-       |                  x$0.asInstanceOf[
-       |                    continuations.compileFromString$package.
-       |                      compileFromString$package$foo$1
-       |                  ]
+       |                $continuation =|                  x$0
```
jackcviers, line 1161: This pipe looks like a typo:

```suggestion
       |                $continuation = x$0
```

jackcviers (changes requested):

jackcviers (changes requested):
Also needs a scalafmtAll

gregorath:
yeah I remember I spent quite some time to make this work like this but annoyingly didn't manage to. it looks like it fails in the `GenBCode` phase. happy to help you debug it if you need to see the tree that fails etc. it would be nice if we manage to do it.

jackcviers (approved):

diesalbla:
An update on this work: yesterday @gregorath and I started looking at the trees generated by the compiler phases. We found the problem to be in the pattern-matcher. Before this phase, the body . However, within the pattern matcher the variable `x$0` is renamed to a fresh variable `x1`, but _only in the pattern_ itself, not in the body of the first `case`.

continuationsPlugin/src/test/scala/continuations/ContinuationsPluginSuite.scala
```diff
@@ -135,19 +135,27 @@ class ContinuationsPluginSuite extends FunSuite, CompilerFixtures, StateMachineF
       }
   }
 
-  compilerContext.test("debug".ignore) {
+  compilerContextWithContinuationsPlugin.test("debug".only) {
```
jackcviers, line 138: You have to remove the .only to get reports on everything in the checks.

jackcviers: We should be able to remove the debug test entirely, we have a better handle on how to get the tree outputs at need now.

diesalbla:
> An update on this work: yesterday @gregorath and I started looking at the trees generated by the compiler phases. We found the problem to be in the pattern-matcher. Before this phase, the body . However, within the pattern matcher the variable x$0 is renamed to a fresh variable x1, but only in the pattern itself, not in the body of the first case.

It seems that by adding an extra wrapper of a `Bind`, with the `Typed` on the right hand side, it now works fine and recognises as a variable (waiting for CI)

PR #92, opened 2023-03-03
No need for  Setters on the continuation variables..

diesalbla:


continuationsPlugin/src/main/scala/continuations/DefDefTransforms.scala
```diff
@@ -780,16 +780,6 @@ object DefDefTransforms extends TreesChecks:
 
     val continuationStateMachineClass: tpd.TypeDef =
 
-      def toContFsmSetter(vd: tpd.ValDef): tpd.DefDef = {
```
gregorath, line 783: from what I remember these setters are actually being added in a different phase from the compiler for all the classes. but this didn't happen for our class, maybe because we need to register/flag it somehow in order to be part of the transformation?
hence we added them ourselves

diesalbla: So, you mean that the code that we are generating is already a few steps into the compiler phases?

gregorath: having the [phases](https://github.com/lampepfl/dotty/blob/main/compiler/src/dotty/tools/dotc/Compiler.scala) our transformation runs after `Staging` and before `PickleQuotes`, can see this in the `ContinuationsPlugin.scala`

the getters/setters transformation happens in the `Memoize` phase but for some reason (maybe we need to "register" our synthetic class to be picked up?) our class was not being transformed to get the setters. that is why we get this error at some point afterwards (can probably search for the error message in dotty to see where it happens). so in the end we added the setters ourselves

gregorath: actually sorry what I said about the `Memoize` bit is not correct. at `Memoize` it transforms the getters/setters. I can't remember if there is a phase that adds them or it always need to be done by whoever created the class. but I copied how the setter should look like from what it expected at `Memoize`

PR #94, opened 2023-03-03
cu-862javy8t: Creates ForceableCompilationPlugin

jackcviers:
The plugin allows you to force compilation with scalac.

It has some limitations.

1. You must have the same version of scala and scalac on your path as is used in the build.sbt. Currently this is 3.1.2.
2. It will only compile scala files.
3. You must run your examples in projects where `ForceableCompilationPlugin` is enabled and `forceCompilation` is `true` with `<projectName> / runMain <fullyQualifiedNameOfMain>`

This is only necessary until https://github.com/sbt/sbt/issues/7157 is resolved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,25 @@
 # scala-fx
 
+## BUILD NOTICE
+
+The following projects have been unaggregated and must be compiled seperately and run with runMain:
+
+* continuationsPluginExample
+* zero-arguments-no-continuation-treeview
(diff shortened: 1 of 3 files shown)


gregorath:
maybe we can add the guidelines from the description in the README file too?

jackcviers:
> maybe we can add the guidelines from the description in the README file too?

Done

PR #95, opened 2023-03-06
Refactor: split continuation symbol out of ValDef.

diesalbla:
- We modify the code generation so that the declaration of the variable `$continuation` and its initialisation, which uses a `match`, all happen in a single statement. 
- We also change the type assigned to the `$continuation` variable, to be an instance of the inner class, so we can save a lot of the `asInstanceOf` statements.

### DIFF
--- a/continuationsPlugin/src/main/scala/continuations/DefDefTransforms.scala
+++ b/continuationsPlugin/src/main/scala/continuations/DefDefTransforms.scala
@@ -853,7 +853,7 @@ object DefDefTransforms extends TreesChecks:
         newSymbol(
           newParent,
           termName("$continuation"),
-          Local | Mutable | Synthetic,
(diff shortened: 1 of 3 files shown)


gregorath (approved):
that's a nice refactoring 👍

gregorath:
we do have the tests but they can't catch everything for now, does it work in runtime for the `continuationPluginExample` project especially the `Main` one which has a lot of cases?

diesalbla:
> we do have the tests but they can't catch everything for now, does it work in runtime for the continuationPluginExample project especially the Main one which has a lot of cases?

I have not tested all cases, but a couple of those seem to be working fine.

gregorath:
okay yes if it works for a couple it should be okay

jackcviers (approved):

PR #99, opened 2023-03-06
InvokeSuspendMethod - the this needs no casting

diesalbla:


gregorath (approved):
it looks good since it works but for the discussion, will the compiler still add this casting at some point later? 
It would be interesting to see how this code would look like after `pickleQuote` phase if it was scala code that went through compilation without our plugin and not something we added during compilation, if that makes sense. like in that case will there be a `asInstanceOf` or not? 
nothing blocking the merge though

PR #101, opened 2023-03-07
DefDefTransform: reduce generated code

diesalbla:
- Change the `Result` type in the `Continuation` companion object to remove the `Either` wrapper.
- Make that type polymorphic. 
- Add an initialisation method to the `SafeContinuation` companion object.


continuationsPlugin/src/test/resources/DefInsideSuspendExpected.txt
```diff
@@ -7,13 +7,10 @@ package continuations {
   () extends Object() { this: continuations.compileFromString$package.type =>
```
gregorath, line 7: could we make this file `.scala` for consistency?

diesalbla: Sure.

diesalbla: done

continuationsPlugin/src/main/scala/continuations/DefDefTransforms.scala
```diff
@@ -1074,7 +1043,7 @@ object DefDefTransforms extends TreesChecks:
         def assignGlobalVarResult(vd: tpd.ValDef): tpd.Assign =
           tpd.Assign(
             globalVars.find(matchesNameCoord(_, vd)).get,
-            ref($result.symbol).select(nme.asInstanceOf_).appliedToType(vd.symbol.info)
+            ref(resultSym).select(nme.asInstanceOf_).appliedToType(vd.symbol.info)
```
gregorath, line 1046: this is good but there is a concern. sometimes there is a difference between referencing the symbol itself vs referencing the symbol after it had been assigned to a tree. I guess it has to do with potential flags it gets or changes in the owners chain.

that is to say if we have tested that everything still works in runtime (compile/tests don't catch all the issues with flags/owners) then yes it should be fine to do this. 
my suggestion would be to run the https://github.com/47deg/TBD/blob/main/runExamples.sh without the cases that throw an exception (see [this](https://github.com/47deg/TBD/pull/81/files#diff-d348e5da10b0a5b51efb46cbf78d239676170c8fa1aa0d36c19786b2a3c0464bR1)) and if everything works (around 50 scenarios) then it is safe

gregorath: unfortunately there are issues atm that we only see in runtime and we don't have such a check in CI. running that script for now is the best we have

gregorath (approved):

PR #105, opened 2023-03-10
CU-8677bmjqw - Compiler Plugin: Suspend in abstract class

gregorath:
for this we had to make sure the `def continuation` signature we create in the transformation is the same for all our cases. 

- ~as a result though this breaks the polymorphic functions in [PolymorphicDefsZeroContinuations.scala](https://github.com/47deg/TBD/blob/main/continuationsPluginExample/src/main/scala/examples/PolymorphicDefsZeroContinuations.scala). this needs to be looked into.~ fixed in [6a997ce](https://github.com/47deg/TBD/pull/105/commits/6a997ce22b703be1d6bea31f39f308aab7d6509e)
-  ~the other issue is calling the method on the point that we create the class eg [new ExampleClass2Impl("AAA").continuations(1, 2)](https://github.com/47deg/TBD/pull/105/files#diff-85be3cab8906bae36ce9cafdd73fc15ab58d848fd47432fe58ff82950b06fc4dR17)~ fixed in [d6d3107](https://github.com/47deg/TBD/pull/105/commits/d6d31075c3e865928e524ed89e59c6672dad8ab9)
- ~existing tests need to be updated~ done in [1391837](https://github.com/47deg/TBD/pull/105/commits/139183704b79259fce8024b2616ff276342a646a)
- new tests need to be added for the abstract class case

going through the commits maybe will help with the review


diesalbla:
The continuations plugin, with its PRs, has been ported to a separate repository.

PR #115, opened 2023-03-16
Example of transformed function as frame class method

diesalbla:


diesalbla:

PR #119, opened 2023-03-17
Avoid use of global vars

diesalbla:
Perhaps we do not need to use the global vars, we can just rely on the fields in the Frame.

_Note_ this could make the compiled code slower. Note, however, that the loading of vars into memory is part of bytecode.


jackcviers:
I am going to pull this down locally and play with it, and see if it is, in fact slower after I'm done calculating overhead.

continuationsPlugin/src/test/resources/StateMachineTwoContinuationsChainedTwoValPriorTwoValBetweenTwoValAfterChainIgnored.scala
```diff
@@ -72,13 +67,11 @@ package continuations {
           $continuation.$label match
             {
               case 0 =>
+                $continuation.I$0 = x
```
gregorath, line 70: I think though that you don't want to do this before you check the result, it seems to me that the behaviour is different now

continuationsPlugin/src/test/resources/StateMachineTwoContinuationsChainedTwoValPriorTwoValBetweenTwoValAfterChainIgnored.scala
```diff
@@ -92,20 +85,11 @@ package continuations {
                       return[label1] ()
                   }
               case 1 =>
-                y##1 = $continuation.I$0
```
gregorath, line 95: how do we track the different values before and after `case 1` then? if it returns before case 1 and then comes back later on to continue from here

gregorath:
one thing to be careful here is that this kind of logic is not fully tested in our code yet. we don't have examples were a continuation is suspended and then resumes after the label has changed because we haven't added this functionality fully yet. or am I missing something?
so for now it is as the decompiled codes told us it should look like

diesalbla:
Since @jackcviers Is reworking the implementation, this is no longer relevant.

PR #122, opened 2023-04-05
Avoid early "flatten"

diesalbla:


diesalbla:

PR #125, opened 2023-04-13
DRAFT: turn "shift" from suspend method into implicit-enabled method.

diesalbla:
Currently, `shift` is a method of `Suspend` so we need to use it with a method call. However, we hardly ever need to refer to the suspend, all we need to do is to restrict that `shift` can only appear within the implicit context of a `Suspend`...

Thus, an implicit `Suspend` grants the capability to `shift`,


diesalbla:
This change is desirable in long run, but will come after current work by @jackcviers and there are other changes to prioritise. Will redo afterwards.

PR #127, opened 2023-04-14
Upgrade Scala, SBT

diesalbla:


build.sbt
```diff
@@ -1,7 +1,7 @@
 import Dependencies.Compile._
 import Dependencies.Test._
 
-ThisBuild / scalaVersion := "3.1.2"
+ThisBuild / scalaVersion := "3.2.2"
```
jackcviers, line 4:
```suggestion
ThisBuild / scalaVersion := "3.1.2"
```
If metals doesn't support it (which I don't think it does as of 11.11).

Then we need to define `crossScalaVersions` for our below projects.

Besides IDE support,  there are tasty incompatibilities between early 3.0 scala libs and late 3.x scala libs that have cropped up in my benchmarking of external libraries,  so we need to compile against the released scala 3 versions and the latest nightly.

If we're upgrading to support broader versions,  might as well do the whole thing. Probably have to update the github actions to use `+` as well.

That could be done in a separate pr(s), however. Thoughts?

diesalbla: Since this is a new library and plugin, perhaps we should upgrade to 3.3.0 and clean the slate then.

PR #128, opened 2023-04-16
Cu 865bt5p1u

jackcviers:


continuationsPluginExample/src/main/scala/examples/PolymorphicDefsZeroContinuations.scala
pabloconde, line 1: Polymorphic functions are not supported with the changes added by async. See bug ticket: https://app.clickup.com/t/865c5yynz

PR #132, opened 2023-05-01
Reworks Launchers/Boundaries to be library-level code

jackcviers:
* Replaces `SuspendApp` with `Blocking`
  `Blocking` executes suspended definitions in sequential order on a single-threaded `ExecutionContext`, returning the resumed values and blocking until the resumed values are available.
  ```scala
  def twoSuspensions(x: Int)(using s:Suspend, ec: ExecutionContext): Int =
    val f = s.shift{ c =>
      ec.execute{() =>
        Thread.sleep(Random.between(1_000, 10_000) 
        c.resume(x + 5)
      }
    }
    s.shift(_.resume(f + 1)
  given executor:ExecutionContextExecutorService = 
    ExecutionContext.fromExecutorService(Executors.newWorkStealingPool())
    
  println(List(1,2,3,4).map(i => Blocking(twoSuspensions(i)))) 
  // List(7, 8, 9, 10), execution of each suspension happens sequentially.
  ```
* Adds `Deferred` Launcher
  `Deferred` launchers launch their suspensions in a non-blocking manner, returning a `Deferred[A]` object with an `await` method to block and extract the `A` value. `IterableOnce[Deferred[A]]`s also have an `awaitAll` that will allow concurrent execution of all of the `Deferred` returns in the `IterableOnce`, but will block the caller until all the `Deferred`s have completed and return a `List[A]`. `awaitAll` is more efficient than `collection.map(i => Deferred(suspendedDef(i)).await()`. 
  
  ```scala
  def twoSuspensions(x: Int)(using s:Suspend, ec: ExecutionContext): Int =
    val f = s.shift{ c =>
      ec.execute{() =>
        Thread.sleep(Random.between(1_000, 10_000) 
        c.resume(x + 5)
      }
    }
    s.shift(_.resume(f + 1)
  given executor:ExecutionContextExecutorService = 
    ExecutionContext.fromExecutorService(Executors.newWorkStealingPool())
    
  println(List(1,2,3,4).map(i => Deferred(twoSuspensions(i))).awaitAll())
  // List(7, 8, 9, 10), execution of each suspension happens concurrently.
  ```
  `Deferred` is not full structured concurrency at this time. There is no cancellation.
  
* [ ] Simplifies the transformation to use simple substitution for `using Suspend`
  Previously, the following suspended definition:
  ```scala
  def suspendedDefinition(x: Int, y: Int)(z: Int)(using Suspend): Int
  ```
  would have its curried arguments flattened in order to insert the necessary `Continuation[Int]` completion parameter:
  ```scala
  def suspendedDefinition(x: Int, y: Int, z: Int, completion: Continuation[Int]): Int
  ```
  This caused some very complicated logic to handle rewrites of applications and the body transformations of the suspended definitions. This necessitated launchers to also be fully transformed, and limited their utility. Now, we simply substitute the completion wherever the contextual `Suspend` parameter appears in the definition, which is both easier to implement and allows launchers to be defined as library-level code rather than as compiler generated code:
  ```scala
  def suspendedDefinition(x: Int, y: Int)(z: Int)(using completion: Continuation[Int]): Int
  ```
  Applications of suspended definitions must still be transformed to include the nearest completion parameter. Launchers all take a `Continuation[A] ?=> A` block in their apply methods. Thus, the parameter can be retrieved from the launchers, which do not need to be transformed and can be implemented simply via substitution. This substitution is only necessary because implicit injection occurs during the previous phases of the compiler, and so must be replicated here:
  ```scala
  // 
  Blocking(suspendedDefinition(1, 2)(3))
  ```
  Becomes (evidence parameter name may differ):
  ```scala
  Blocking(suspendedDefinition(1, 2)(ev$1)
  ```
  This prevents many issues with ownership and proxying due to the previous complex transformations.

* [ ] Environment Variable capture is now performed with a `deepFold` and `filter` instead of takeWhiles and complex `fold`s.
  
* [ ] State machine generation now operates almost solely as `TreeTypeMap`s. This makes generating state machine cases a lot less complex. Future improvements could be made to consolidate the mapping into fewer `TreeTypeMap`s.

* [ ] Constant definition names have been extracted into `constants.scala`. This will allow the transformation to be  parameterized in the future.

* [ ] Frame class generation has been inlined into the main `TreeTypeMap` transformation, and names have been clarified. `Iterator`s have been used to generate the names from the captured  environment values. 

* Suspended `ValDef` declarations no longer work. A follow-up PR will address this issue in a manner similar to the `DefDef` transformations. `ValDef` contextual function transformation is quite different from `DefDef` transformation, due to access to parameters provided by `DefDef` and symbols with a guaranteed `MethodType` type.


continuationsPlugin/src/main/scala/continuations/ContinuationsFrameBuilder.scala
```diff
@@ -0,0 +1,189 @@
...
+          newSymbol(
+            continuationFrameSymbol,
+            value.symbol.asTerm.name.asTermName.setterName,
+            Flags.Accessor | Flags.Method,
+            MethodType(List(value.symbol.asTerm.info), unitType)
+          ).entered.asTerm,
+          params =>
+            Assign(This(continuationFrameSymbol).select(value.name), ref(params(0)(0).symbol))
+        )
+      } ++ List(
```
diesalbla, line 103:
```suggestion
    def setterDefDef(value: ValDef): DefDef = 
      DefDef(
        newSymbol(
          continuationFrameSymbol,
          value.symbol.asTerm.name.asTermName.setterName,
          Flags.Accessor | Flags.Method,
          MethodType(List(value.symbol.asTerm.info), unitType)
        ).entered.asTerm,
        params =>
          Assign(This(continuationFrameSymbol).select(value.name), ref(params(0)(0).symbol))
      )

    ClassDefWithParents(
      continuationFrameSymbol,
      frameClassConstructor,
      List(
        New(requiredClassRef(continuationImplFullName))
          .select(nme.CONSTRUCTOR)
          .appliedToArgss(List(List(
            $completionRef,
            $completionRef.select(contextName.sliceToTermName(0, contextName.size)))))),
      bodyInputVars ++ bodyInputVars.map(setterDefDef) ++ List(
```

diesalbla:
The compiler plugin has been extracted to a separate repository.

PR #133, opened 2023-05-23
Continuation Plugin: add GCD example.

diesalbla:
Simple example of recursive coroutine which calls other coroutines.


diesalbla:
We are porting the compiler plugin code to a different repository.

